PR #2, opened 2025-05-15
File creation and LiteLLM base

colinfrisch:
This is simply the file creation for our LLMAgent as well as a working LiteLLM integration.

### DIFF
--- a/mesa_llm/module_llm.py
+++ b/mesa_llm/module_llm.py
@@ -1,0 +1,36 @@
+from litellm import completion
+import os
+
+
+
+class ModuleLLM:
+    """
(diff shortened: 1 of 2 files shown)


sanika-n (approved):
Thanks Colin! The LiteLLM was a great find :) 
The code looks great to me:)

mesa_llm/module_llm.py
```diff
@@ -0,0 +1,34 @@
+import os
+
+from litellm import completion
```
wang-boyu, line 3: might want to add `litellm` dependency into our `pyproject.toml`

colinfrisch: Done !

PR #3, opened 2025-05-16
add a class for managing tools

sanika-n:
### The Issue
In the reasoning.py module (for in-context-learning methods) we will require the user to list out their tools following the schema below:
image
(The above image shows the structure while using liteLLM)
Since this can be quite painful, to repeat for each and every tool, I thought having a Tools manager class would be useful.

###  Functionality
- The user can use insntaces of ToolManager to regsiter functions as tools through the decorator (as shown in the example below).
- The user can also use the ToolManager instance to get the schema of the tools, call a tool, and check if a tool is registered.
- Moreover, the user can group like tools together by creating a new ToolManager instance and registering the tools to it. 
So if agent A requires tools A1, A2, and A3, and agent B requires tools B1, B2, and B3, the user can create two ToolManager instances: tool_manager_A and tool_manager_B.

### Example
Code:
```python
tool_manager = ToolManager()
@tool_manager.register
def add_two_numbers(a: int, b: int) -> str:
    """Takes two numbers and sums them up"""
    c = a + b
    return json.dumps({"a": a, "b": b, "c": c})
@tool_manager.register
def multiply_two_numbers(x: int, y: int) -> str:
    """Takes two numbers and multiplies them"""
    z = x * y
    return json.dumps({"x": x, "y": y, "z": z})

print(tool_manager.get_schema())
print(tool_manager.call("add_two_numbers", {"a": 1, "b": 2}))
print(tool_manager.has_tool("add_two_numbers"))
```
Output:
image

### DIFF
--- a/mesa_llm/tools/__init__.py
+++ b/mesa_llm/tools/__init__.py
@@ -1,2 +1,3 @@
 from .tool_manager import ToolManager
+
 __all__ = ["ToolManager"]
--- a/mesa_llm/tools/__init__.py
+++ b/mesa_llm/tools/__init__.py
@@ -1,0 +1,2 @@
+from .tool_manager import ToolManager
+__all__ = ["ToolManager"]
--- a/mesa_llm/tools/inbuilt_tools.py
+++ b/mesa_llm/tools/inbuilt_tools.py
@@ -1,2 +1,0 @@
-from .tool_manager import ToolManager
-__all__ = ["ToolManager"]
--- a/mesa_llm/tools/inbuilt_tools.py
+++ b/mesa_llm/tools/inbuilt_tools.py
@@ -1,0 +1,2 @@
+from .tool_manager import ToolManager
+__all__ = ["ToolManager"]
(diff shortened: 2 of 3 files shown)


colinfrisch (approved):
Decorator is a great idea, and it's probably going to save a lot of time to the user. I'm approving and leaving the merge to @sanika-n or @wang-boyu

wang-boyu (approved):

wang-boyu:
LGTM! Merging.

Would it be useful to have some metadata for tools (e.g., name and description of tools) for LLMs, or are they redundant?

sanika-n:
Thanks! Yes, the LLMs will be accessing the tools through the schema, and I’ve included the name and description there.
Would you like me to store this metadata elsewhere as well, for example, in a dictionary inside the ToolManager class?

wang-boyu:
No worries! Sorry I must have missed the schema output and didn't see how name and description were already added. Was wondering about a more natural language-like function name such as 'add two numbers' vs 'add_two_numbers' in python, but this probably doesn't matter. Or maybe passing the python function names directly is even better. Anyways it's all good!

PR #11, opened 2025-06-16
Memory refactor

colinfrisch:
## Big refactor of the memory 

Before this refactor, each (observation, tool call, plan, etc.) event was recorded on the spot. This made memory a big disorganised pile of event. Displaying it was a pain, and it hindered LLM interpretation.

**Memory is now recorded by steps !** :  during the agent's step, each event is sent to the memory, which stores it safely until the end of the step. When the step is finished, all events are "flushed" and organised into a `MemoryEntry` object that is stored in the short term memory. This also allows to remove all external displaying configuration for a step, since it's already processed by memory.

### What changed ?

- A single memory entry now represents all the content gathered by an agent during a step (in a `content` dictionary which is easy to iterate through).

```python
@dataclass
class MemoryEntry:
    content: dict
    step: int

"""
content is a dict that can have nested dictionaries of arbitrary depth of the form :
{
    "Observation": {
        "self_state": {
            "system_prompt": "You are an agent in a simulation...",
            "location": "1,1",
            "internal_state": "..."
        },
        "local_state": {
            "Agent1": {"position": "1,1", "internal_state": "..."},
            "Agent2": {"position": "2,2", "internal_state": "..."}
        }
    },

    "Plan": {
        "step": 1,
        "llm_plan": "...",
    },

    "Tool_call_response": {
        "tool": "speak_to",
        "content": "..."
    }
}

"""
```

- No more external step displaying (removed `agent_step_display.py)` : an agent step (or its corresponding `entry = MemoryEntry`) is styled with the internal `entry.style_format()` method. This allowed to removed a few small not-so-used methods in Memory.

### But wait ! There is more !

Working on this also made me do a few modifications of the core codebase for more efficiency : 
- Enforced LLM output in ReAct with the `ReActOutput` data class (reasoning : str, action : str). It's passed as an `output_format` parameter in the generate method.
- made new method for function calling from a text input (basically separated the code that was already written on this to put it in the more global reasoning class for clarity purposes)


sanika-n:
Thanks! This makes it a lot cleaner... Pydantic is also a great move:)

PR #13, opened 2025-06-19
Enhanced Negotiation

sanika-n:
Just added some more elements into negotiation. Have pushed a readme file as well just to elaborate on the changes.

### DIFF
--- a/examples/negotiation/model.py
+++ b/examples/negotiation/model.py
@@ -44,13 +44,31 @@ class NegotiationModel(Model):
 
         agents = BuyerAgent.create_agents(
             self,
-            n=initial_buyers,
+            n=initial_buyers - math.floor(initial_buyers / 2),
             api_key=api_key,
             reasoning=reasoning,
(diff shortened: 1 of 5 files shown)


examples/negotiation/model.py
```diff
@@ -44,13 +44,31 @@ def __init__(
 
         agents = BuyerAgent.create_agents(
             self,
-            n=initial_buyers,
+            n=initial_buyers - math.floor(initial_buyers / 2),
```
colinfrisch, line 47: I don't really get the point of this line : if you want half of them couldn't you do it without the subs traction ?

colinfrisch (approved):

colinfrisch:
All good for me ! just that small interrogation regarding the math line and we can merge :)

examples/negotiation/model.py
colinfrisch: Okay, thanks for the explanation :)

PR #14, opened 2025-07-01
Add parallel stepping to mesa-llm

colinfrisch:
# Add parallel stepping to mesa-llm

## General idea

Allows to run llm-based simulations faster by making steps for agents simultaneous for each model step by using `async` modules in python. The initial idea was to implement a batching mechanism for LLM calls, but litellm structure makes it so that the way that batching is implemented is not aligned with mesa-llm objectives.

> **IMPORTANT :** since all steps are executed in parallel batching is not compatible with synchronous events such as conversations.

## Additions to mesa-llm

Synchronous step() agents are wrapped to fit the same interface. 

- `parallel_stepping.py` : where the async stepping is defined and generalized. Also introduces a wrapper to AgentSet `suffle_do` method so that it doesn't change anything when calling model step.
- Add async plan to all reasoning methods

- Refactoring changes to reasoning methods : add a `get_react_system_prompt()`, `get_cot_system_prompt()`, etc. to all reasoning methods to avoid too much redundancies with async `plan()` method

- Example model "random bunnies" : just agents (bunnies) randomly hopping (moving on the grid) and that's all

## How to use

- Set `self.parallel_stepping = True` in the model’s __init__()
- In agents file, make an async function astep() with the code content of step like so (extracted from random bunnies example): 

```python
    async def astep(self):
        """
        Step the bunny agent (asynchronous version).

        Note :
        """
        prompt = "Move around the grid."
        plan = await self.reasoning.aplan(
            prompt=prompt,
            obs=self.generate_obs(),
            selected_tools=["move_one_step", "teleport_to_location"],
        )
        self.apply_plan(plan)
        print(f"Bunny {self.unique_id} moved to {self.pos}")
```

> In this code, don't forget to call planning with `plan = await self.reasoning.aplan`


mesa_llm/module_llm.py
```diff
@@ -30,34 +37,57 @@ def __init__(self, api_key: str, llm_model: str, system_prompt: str | None = Non
...
                 {"role": "user", "content": prompt},
             ]
-        else:
-            messages = [{"role": "user", "content": prompt}]
-        if tool_schema:
+            if self.system_prompt
+            else [{"role": "user", "content": prompt}]
+        )
+
+        if self.batch_size == 1:
```
sanika-n, line 56: Are you planning on extending the code later to make batching possible? If not, I think we can remove self.batch_size right? As currently I see that only batch_size==1 (i.e. no batching case ) has been implemented

mesa_llm/llm_agent.py
```diff
@@ -225,30 +225,74 @@ def pre_step(self):
...
         This is some code that is executed after the step method of the child agent is called.
         It functions because of the __init_subclass__ method that creates a wrapper around the step method of the child agent.
         """
         self.memory.process_step()
 
+    async def apost_step(self):
+        """
+        Async version of post_step() for asynchronous stepping.
+        """
+        self.memory.process_step()
```
sanika-n, line 245: Is there a need to make apre_step and apost_step async?

sanika-n (commented):
Overall looks great!! Really like the fact that creating a parallel stepping model is not vastly different from creating a normal model:) 
Had a few questions, but once resolved I think it is ready to be merged!

mesa_llm/module_llm.py
colinfrisch: Yes, that's a residue from experimentations, well spotted :) 
Will delete in next commit

mesa_llm/llm_agent.py
colinfrisch: Just tested without and it's the same perf ! 
Removed them in commit.

PR #78, opened 2026-02-20
Fix STLTMemory consolidation to respect consolidation_capacity batch size

### LINKED ISSUE #77: STLTMemory ignores consolidation_capacity batch size
**Describe the bug**

`STLTMemory `does not honor `consolidation_capacity` as a consolidation batch size, the current consolidation behavior effectively treats it like a trigger threshold and removes only one entry per consolidation cycle instead of removing `consolidation_capacity` entries.

**Expected behavior**

When Short-Term Memory (STM) exceeds the configured capacity, the memory system should remove and consolidate `consolidation_capacity` entries at a time (e.g., if `consolidation_capacity=3`, then 3 oldest entries should be removed from STM and consolidated into Long-Term Memory (LTM) in a single consolidation cycle).

**To Reproduce**

Create an STLTMemory instance with a small short-term capacity and a larger consolidation capacity, e.g.:

1. short_term_capacity = 2
2. consolidation_capacity = 3
3. Insert enough entries so STM exceeds the effective limit (e.g., add 6 total entries).
4. Trigger consolidation (via the normal process_step flow).
5. Observe the STM size after consolidation.

**Example (simplified test scenario):**

- Start with 6 entries in STM
- Expected: 6 - 3 = 3 entries remain after one consolidation cycle
- Actual: only 6 - 1 = 5 entries remain after one consolidation cycle

**Additional context**

This behavior can keep STM in an “overflow” state for multiple steps, leading to inefficient memory management.

psbuilds:
### Summary

This PR fixes an issue in `STLTMemory` where `consolidation_capacity` was not properly respected as a batch size during consolidation. Previously, consolidation removed only a single entry per cycle, even when consolidation_capacity was configured to remove multiple entries.

This resulted in inefficient short-term memory cleanup and inconsistent behavior relative to the documented design.

### Bug / Issue
Fixes #77 

The `STLTMemory` class documentation states that `consolidation_capacity` controls the number of entries to summarize when short-term memory exceeds its effective limit.

**Expected behavior:**
When STM size exceeds` short_term_capacity + consolidation_capacity`, the memory system should remove and consolidate `consolidation_capacity` entries at once.

**Actual behavior:**
Only one entry was being removed during each consolidation cycle, effectively treating consolidation_capacity as a threshold trigger rather than a batch size.

This caused:
- STM to remain in an overflow state for multiple steps
- Consolidation to be slower and less efficient
- Behavior inconsistent with documentation

### Implementation

Changes made:

- Updated the consolidation logic inside _process_step_core in STLTMemory.
- Modified the removal behavior to correctly pop consolidation_capacity entries using a loop:
- Ensured consolidation is triggered only when:
          `len(self.short_term_memory) > self.capacity + self.consolidation_capacity`
- Added a unit test (`test_batch_size_fix`) to explicitly verify correct batch removal behavior.

### Additional Notes

- This change improves memory management efficiency.

### DIFF
--- a/mesa_llm/memory/st_lt_memory.py
+++ b/mesa_llm/memory/st_lt_memory.py
@@ -134,7 +134,8 @@ class STLTMemory(Memory):
             > self.capacity + (self.consolidation_capacity or 0)
             and self.consolidation_capacity
         ):
-            self.short_term_memory.popleft()
+            for _ in range(self.consolidation_capacity):
+                self.short_term_memory.popleft()
             should_consolidate = True
 
         elif (
--- a/tests/test_memory/test_STLT_memory.py
+++ b/tests/test_memory/test_STLT_memory.py
@@ -1,5 +1,5 @@
 from collections import deque
-from unittest.mock import patch
+from unittest.mock import Mock, patch
 
 from mesa_llm.memory.memory import MemoryEntry
 from mesa_llm.memory.st_lt_memory import STLTMemory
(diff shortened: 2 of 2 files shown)


psbuilds:
hey @sanika-n any updates on this :)

psbuilds:
> Hey, a bit tied up at the moment, I have requested Colin to take a look, in case he is busy as well, I will be able to check it out in a couple of days

Sure! Thanks a lot :)

psbuilds:
Hey @sanika-n happy to help with this further after discussions with @colinfrisch if you do need any sort of help :)

PR #109, opened 2026-03-01
fix: store graded EpisodicMemory entries as MemoryEntry objects and use correct llm instance 

### LINKED ISSUE #108: EpisodicMemory: Graded entries are never stored, as a result LLM never receives memory context
**Describe the bug**
EpisodicMemory has an `add_to_memory()` function that grades each event's importance, but the graded content is never stored as a MemoryEntry in `self.memory_entries`.

 The grading result is placed into `self.step_content` (via `super().add_to_memory()`), which is then immediately cleared in `process_step`. 

As a result, `memory_entries` is always empty, `retrieve_top_k_entries()` returns nothing, and `get_prompt_ready()` provides no memory context to the LLM during reasoning.

**Expected behavior**

After calling `add_to_memory()`, the graded event should be stored as a MemoryEntry in `self.memory_entries` so that:

- `retrieve_top_k_entries()` can retrieve the most important + recent memories
- `get_prompt_ready()` returns meaningful context to the LLM during reasoning

**To Reproduce**
```
from mesa_llm.memory.episodic_memory import EpisodicMemory
import json

agent = MagicMock()
agent.model.steps = 1
mock_response = MagicMock()
mock_response.choices[0].message.content = json.dumps({"grade": 4})
agent.llm.generate.return_value = mock_response

memory = EpisodicMemory(agent=agent, llm_model="provider/test_model")

# Add a graded memory
memory.add_to_memory("observation", {"data": "saw 3 cops nearby"})

# Bug: memory_entries is empty even after adding a graded event
print(len(memory.memory_entries))  # → 0 (expected: 1)
print(memory.get_prompt_ready())   # → "Top 5 memory entries:\n\n" (no actual entries)
```

**Additional context**
I was initially going through the logic for grading in `episodic_memory` which also seems to be flawed when a lot of steps are involved in the agents simulation.

Due to the lack of normalisation over long simulations the grading logic seems to give more weight to recency rather than the other two factors, when in reality according to the research paper mentioned in the file all the factors must be given equal weight.

But this can be tackled only after fixing the bug mentioned in this issue as without entires to grade and supply to the LLM properly `episodic_memory` file doesn't make sense.

If I am wrong in my interpretation please do help me to understand where I went wrong in interpreting this. :)

I'm currently working on a PR for this issue

psbuilds:
### Summary
In EpisodicMemory graded entries were calculated but never stored as `MemoryEntry` objects, causing `get_prompt_ready()` to return empty memory context to the LLM. Also fixed the grading methods using the wrong LLM instance.

### Bug / Issue
Fixes: #108 

**add_to_memory never creates MemoryEntry objects:** 

`add_to_memory()` grades each event via an LLM call, but only stores the result in step_content (via `super().add_to_memory()`), which is immediately cleared. No MemoryEntry is ever created, so memory_entries stays empty and `retrieve_top_k_entries() / get_prompt_ready()` return nothing to the LLM.

**Wrong LLM instance for grading:** 
`grade_event_importance()` sets the system prompt on `self.llm` but calls `self.agent.llm.generate()`. This means the grading system prompt is never used and the agent's own system prompt gets silently overwritten during grading.

### Implementation

**episodic_memory.py:**
`add_to_memory()`/ `aadd_to_memory()` : Now creates a MemoryEntry with the graded content and appends it to memory_entries. Uses `{**content, "importance": grade}` to avoid mutating the input dict.
`grade_event_importance() / agrade_event_importance()` : Changed `self.agent.llm.generate() → self.llm.generate()` so the grading system prompt is actually used.

### Testing
The already existing test files were updated to prove this behaviour fix.


wang-boyu:
We can probably do it together in this PR.

Two questions -

- Is `importance` a key in memory entry content, or one level below inside `content[content_type]`? I'm not sure whether this works as expected:
   ```python
   key=lambda x: x.content["importance"] - (self.agent.model.steps - x.step)
   ```
- I guess we were tring to use `self.agent.model.steps - x.step` as some sort of `recency` measure? The paper uses
  * `recency` score that decays exponentially with a factor of 0.995
  * `importance` score that ranges from 1 to 10, later min-max scaled to `[0, 1]`
  * there's also a `relevance` score that changes according to an incoming query
 
   Then the final overall ranking (called "retrieval score") is computed as the sum of these three. They are added as a weighted sum controlled by some $\alpha$ coefficients, but they are all set to 1 so it's just a simple sum.

   I suppose the purpose of the min-max scaling on `importance` is to match the range of other scores, so that it does not outweigh them.

wang-boyu (approved):

PR #118, opened 2026-03-01
fix: align STLTMemory.get_prompt_ready() return type with Memory ABC

### LINKED ISSUE #113: ReActReasoning assumes list-based memory prompt and breaks with non-STLT memory backends

khushiiagrawal:
### Summary

`STLTMemory.get_prompt_ready()` returned a `list` instead of `str`, violating the `Memory` ABC contract. Fixed to return a joined string and updated `ReActReasoning` to handle the corrected return type.

### Bug / Issue

Fixes #113

`STLTMemory.get_prompt_ready()` returned a `list[str]` while the `Memory` ABC declares `-> str`. All other backends (`ShortTermMemory`, `LongTermMemory`, `EpisodicMemory`) correctly return `str`. Since `STLTMemory` is the default memory in `LLMAgent`.

### Implementation

**`mesa_llm/memory/st_lt_memory.py`**
- Joined the two list items into a single string with `"\n\n"` separator

**`mesa_llm/reasoning/react.py`**
- `get_react_prompt()` now wraps the `str` result from `get_prompt_ready()` in a list, since it later appends communication history and observation items

**`tests/test_reasoning/test_react.py`**
- Updated 7 mock return values from `list` to `str`

**`tests/test_memory/test_STLT_memory.py`**
- Added 2 regression tests asserting `get_prompt_ready()` returns `str` (with populated and empty memory)

### Testing

All tests pass
Screenshot 2026-03-01 at 6 51 20 PM

Screenshot 2026-03-01 at 6 51 14 PM

### DIFF
--- a/mesa_llm/memory/st_lt_memory.py
+++ b/mesa_llm/memory/st_lt_memory.py
@@ -193,10 +193,10 @@ class STLTMemory(Memory):
             return "\n".join(lines)
 
     def get_prompt_ready(self) -> str:
-        return [
-            f"Short term memory:\n {self.format_short_term()}",
-            f"Long term memory: \n{self.format_long_term()}",
-        ]
+        return (
+            f"Short term memory:\n {self.format_short_term()}\n\n"
+            f"Long term memory: \n{self.format_long_term()}"
+        )
 
     def get_communication_history(self) -> str:
         """
--- a/mesa_llm/reasoning/react.py
+++ b/mesa_llm/reasoning/react.py
@@ -46,7 +46,7 @@ class ReActReasoning(Reasoning):
         return system_prompt
 
     def get_react_prompt(self, obs: Observation) -> list[str]:
-        prompt_list = self.agent.memory.get_prompt_ready()
+        prompt_list = [self.agent.memory.get_prompt_ready()]
         last_communication = self.agent.memory.get_communication_history()
 
         if last_communication:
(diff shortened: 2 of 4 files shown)


khushiiagrawal:
hii @wang-boyu , happy to contribute to mesa/llm , could you please take a look.
thanks!

wang-boyu (approved):

PR #133, opened 2026-03-02
fix: use agenerate_obs in ReWOOReasoning.aplan() instead of blocking generate_obs

### LINKED ISSUE #132: bug: ReWOOReasoning.aplan() uses sync generate_obs() instead of async
**Describe the bug**

`ReWOOReasoning.aplan()` calls `self.agent.generate_obs()` (sync) instead of 
`await self.agent.agenerate_obs()` on line 171 of `rewoo.py`. Since this is inside 
an async method, it blocks the event loop on every planning step — making async 
parallel execution behave exactly like sync.

**Expected behavior**

`aplan()` should use the async version:

self.current_obs = await self.agent.agenerate_obs()

`agenerate_obs()` already exists on `LLMAgent` for this purpose.

**To Reproduce**

Use any agent with `ReWOOReasoning` in async/parallel stepping mode 
(`step_agents_parallel` or `do_async`). The observation step will silently 
block the event loop instead of running asynchronously.

**Additional context**

The sync `plan()` method is fine — only `aplan()` has this issue. 
One-line fix: replace `self.agent.generate_obs()` with 
`await self.agent.agenerate_obs()` on line 171.

khansalman12:
Fixes #132

ReWOOReasoning.aplan() was calling self.agent.generate_obs() (synchronous) instead of await self.agent.agenerate_obs(). This blocked the event loop on every planning step when using async parallel stepping, making async execution behave like sync.

**Changes:**
- `rewoo.py`: replaced `self.agent.generate_obs()` with `await self.agent.agenerate_obs()` in `aplan()`.                

- - `test_rewoo.py`: updated affected tests to mock `agenerate_obs` (AsyncMock) instead of `generate_obs`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ ci:
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.14.14
+  rev: v0.15.4
   hooks:
     # Run the linter with fix argument.
     - id: ruff-check
--- a/mesa_llm/memory/memory.py
+++ b/mesa_llm/memory/memory.py
@@ -137,6 +137,12 @@ class Memory(ABC):
         """
         Add a new entry to the memory
         """
+        if not isinstance(content, dict):
+            raise TypeError(
+                "Expected 'content' to be dict, "
+                f"got {content.__class__.__name__}: {content!r}"
+            )
+
         if type == "observation":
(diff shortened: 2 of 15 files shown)


wang-boyu:
I made some unnecessary fix for this issue as part of #131. Merged `main` back into this PR branch and uses your fix instead.

Merging now.

wang-boyu (approved):

PR #145, opened 2026-03-04
fix: handle non-LLM neighbors in _build_observation

### LINKED ISSUE #143: `_build_observation` crashes with AttributeError when a non-LLM agent is a neighbor
**Describe the bug**
When `LLMAgent` has `vision=-1` (or any positive vision radius), it collects nearby
agents and builds a `local_state` dict in `_build_observation()`. This method
assumes every agent in the simulation has an `internal_state` attribute:
```python
"internal_state": [
    s for s in i.internal_state if not s.startswith("_")
],
```

`internal_state` is only set by `LLMAgent.__init__`. Any plain Mesa `Agent`
subclass (e.g. rule-based cops) will cause:

```
AttributeError: 'PlainAgent' object has no attribute 'internal_state'
```
This matters in mixed-agent simulations where some agents are LLM-driven and others follow
simple rules. The Epstein Civil Violence example in this repo is also in this pattern. Right now,
enabling `vision` in any such simulation silently breaks observation generation.

**Expected behavior**
`generate_obs()` should treat agents without `internal_state` as having an
empty internal state, the same way `position` already uses `getattr(i, "cell", None)`
defensively.

**To Reproduce**
```python
import os
from mesa.agent import Agent
from mesa.model import Model
from mesa_llm.reasoning.react import ReActReasoning

os.environ["GEMINI_API_KEY"] = "just checking"

class PlainAgent(Agent):
    def step(self): pass

class MyModel(Model):
    def __init__(self):
        super().__init__(seed=42)
        self.grid = MultiGrid(5, 5, torus=False)

model = MyModel()
llm = LLMAgent(model=model, reasoning=ReActReasoning, vision=-1)
plain = PlainAgent(model=model)
model.grid.place_agent(llm, (2, 2))
model.grid.place_agent(plain, (3, 3))
llm.memory.add_to_memory = lambda *a, **kw: None
llm.generate_obs() 
```
**Additional context**
I will be raising a PR for this soon.

BhoomiAgrawal12:
### Summary
`_build_observation()` in `llm_agent.py` builds each neighbor's `local_state`
entry by reading `i.internal_state` directly. This works fine as long as every
agent is an `LLMAgent`, but throws `AttributeError` the moment a plain Mesa
`Agent` ends up in the neighbors list  which happens in any mixed-agent
simulation with vision enabled.

This PR replaces the bare attribute access with a safe `getattr(..., [])` call so
non-LLM neighbors are silently given an empty internal state instead of crashing
the whole observation step.
### Bug / Issue
Closes #143 

### Implementation
**Before**
```
s for s in i.internal_state if not s.startswith("_")
```
 **After**
```
 s for s in getattr(i, "internal_state", []) if not s.startswith("_")
```
`getattr(i, "internal_state", [])` returns `[]` when the attribute is absent, so
non-LLM neighbors produce an empty internal state instead of a crash. This is
consistent with how `position` in the same loop already uses
`getattr(i, "cell", None)` defensively for the same reason.

### Testing
Two regression tests added to test_llm_agent.py one for each call path into `_build_observation()`
Both tests fail on main before this fix and pass after it.

### Additional Notes
The Epstein Civil Violence example (examples/epstein_civil_violence/) is a real-world instance of this exact pattern, LLM agents coexisting with non-LLM agents. This fix unblocks any mixed-agent simulation from using `vision`.


colinfrisch (approved):
great, thank you !

PR #147, opened 2026-03-05
Fix/ handle different LLM encodings for listener IDs in speak_to

gamal1osama:
**Description**

The `speak_to` tool expects `listener_agents_unique_ids` to be a `list[int]`, but LLMs often send it in different formats depending on how they serialize arguments:

- As a **stringified list**: `"[5, 9]"` (the LLM JSON-encodes it twice by mistake)
- As a **bare integer**: `5` (when there's only one recipient)
- As **None**: happens occasionaly when the LLM forgets the arg

Without this fix, the code raises a `TypeError` (e.g. `'in <string>' requires string as left operand, not int`) which gets caught by `ToolManager._process_tool_call` and returned to the LLM as a cryptic `"Error: ..."` — the message is never delivered and the LLM has no useful feedback to recover from.

This PR adds normalization logic to handle all three cases, so messages get through regaurdless of how the LLM formats the IDs.

**Example scenarios that now work:**

```python
# LLM sends a stringified list (common in chain-of-thought outputs)
speak_to(agent, "[11, 12]", "hey")  # now works, was TypeError before

# LLM sends a bare int (when generating single-recipient calls)
speak_to(agent, 11, "hey")  # now works, was TypeError before

# LLM sends None
speak_to(agent, None, "hey")  # now gracefully handles it
```

**Testing**

Added 5 new tests covering every normalization branch:

| Test | What it covers |
|---|---|
| `test_speak_to_accepts_stringified_list` | `"[11, 12]"` parsed via `ast.literal_eval` |
| `test_speak_to_accepts_bare_int` | `11` wrapped to `[11]` |
| `test_speak_to_handles_none_ids` | `None` treated as empty list |
| `test_speak_to_handles_malformed_string_ids` | garbage string falls back to `[]` |
| `test_speak_to_handles_non_numeric_ids_in_list` | `["alice", "bob"]` → `int()` fails → `[]` |

The first two fail without this PR (raise `TypeError`), the last three cover the except/fallback branches for full patch coverage. All existing tests still pass.

### DIFF
--- a/mesa_llm/tools/inbuilt_tools.py
+++ b/mesa_llm/tools/inbuilt_tools.py
@@ -1,3 +1,4 @@
+import ast
 from typing import TYPE_CHECKING, Any
 
 from mesa.discrete_space import (
@@ -137,6 +138,27 @@ def speak_to(
         listener_agents_unique_ids: The unique ids of the agents receiving the message
         message: The message to send
     """
+
+    # Normalize: the LLM may pass IDs as a JSON string (e.g. '[5, 9]') or a
+    # bare int rather than a proper list[int], which would silently produce
+    # an empty recipient list.
+    if isinstance(listener_agents_unique_ids, str):
+        try:
+            listener_agents_unique_ids = ast.literal_eval(listener_agents_unique_ids)
+        except Exception:
+            listener_agents_unique_ids = []
+
+    if isinstance(listener_agents_unique_ids, int):
+        listener_agents_unique_ids = [listener_agents_unique_ids]
+
+    if listener_agents_unique_ids is None:
+        listener_agents_unique_ids = []
+
+    try:
+        listener_agents_unique_ids = [int(uid) for uid in listener_agents_unique_ids]
+    except Exception:
+        listener_agents_unique_ids = []
+
     listener_agents = [
         listener_agent
         for listener_agent in agent.model.agents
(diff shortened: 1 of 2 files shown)


gamal1osama:
Thanks @colinfrisch  for the feedback. To clarify — the response_format=ReActOutput https://github.com/mesa/mesa-llm/blob/a719dac14e3f314a29cf8758114ba349e99fbb29/mesa_llm/reasoning/react.py#L94
(Pydantic) protects the reasoning output (the first LLM call with tool_choice="none"). 
But the speak_to type issue happens in the tool execution call (the second LLM call with tool_choice="required"), where tool arguments come as raw JSON strings parsed by json.loads() in ToolManager._process_tool_call(). Pydantic doesn't cover that layer.

For example, in cot.py line 135: rsp = llm.generate(prompt=chaining_message, tool_schema=..., tool_choice="required") 
https://github.com/mesa/mesa-llm/blob/a719dac14e3f314a29cf8758114ba349e99fbb29/mesa_llm/reasoning/cot.py#L135

no response_format here. The LLM can still send listener_agents_unique_ids as "[5, 9]" (string) or 5 (bare int) instead of [5, 9] (list).

I'm currently deep-diving into the codebase as part of working on my GSoC proposal, and I found this while tracing the full cognitive loop. Happy to discuss further or reopen if you'd like!
and may be iam wrong with any misunderstanding or missing any edgecases, that is welcome to clarify to me

PR #149, opened 2026-03-05
ToolManager.get_all_tools_schema() exposes every tool to the LLM regardless of agent state.

### LINKED ISSUE #148: `ToolManager.get_all_tools_schema()` exposes every tool to the LLM regardless of agent state
`ToolManager.get_all_tools_schema()` exposes every tool to the LLM regardless of agent state.

A wolf with `0` energy still sees `move_one_step`.  
An agent with no grid still sees `teleport_to_location`.

The LLM picks from all options — if it picks wrong, you get either a crash or a silent bad state.

### The Core Problem

Right now in `ToolManager.get_all_tools_schema()`:

```python
# Every agent, every step, gets EVERY tool
return [fn.__tool_schema__ for fn in self.tools.values()]
```

A wolf with `0` energy still sees `move_one_step`.  
An agent at the grid boundary still sees `teleport_to_location` with no hint that `(5, 99)` is out of bounds.

The LLM selects from **all possible tools** and hopes the choice is valid.

This ignores whether the tool is actually usable in the current state.

### Example Issues

| Tool | Implicit Assumption | What Breaks |
|-----|-----|-----|
| `move_one_step` | `agent.pos` is not `None` | `TypeError: cannot unpack non-iterable NoneType` |
| `move_one_step` | `grid/space` exists on model | `AttributeError` |
| `teleport_to_location` | coordinates in bounds | silent wrong state or `IndexError` |
| `speak_to` | target agents exist | empty list → silent no-op |

The LLM selects from **all possible tools**, even when they are impossible to execute.

---

## Proposed API

Keep the existing method and add two new ones.

```python
class ToolManager:

    # Existing (unchanged)
    def get_all_tools_schema(self, selected_tools=None) -> list[dict]: ...

    # NEW — hard filter (executor safe)
    def get_feasible_tools_schema(self, agent, selected_tools=None) -> list[dict]: ...

    # NEW — soft annotations with reasons (planner guidance)
    def get_annotated_tools_schema(self, agent, selected_tools=None) -> list[dict]: ...
```

### Pros

- Clear API
- Explicit semantics
- No breaking changes
- Executors only receive feasible tools
- Planners can still reason about infeasible tools

---

## Files That Change

| File | Change |
|-----|-----|
| `tool_decorator.py` | Add `@requires` decorator |
| `tool_manager.py` | Add `get_feasible_tools_schema()` and `get_annotated_tools_schema()` |
| `llm_agent.py` | Add `tool_filter()` method |
| `inbuilt_tools.py` | Add `@requires` to built-in tools |
| `cot.py` | Annotated for planner, feasible for executor |
| `react.py` | Annotated for planner |
| `rewoo.py` | Annotated for planner |
| `reasoning.py` | Use feasible tools for execution |

hillhack:
## Fix
Closes: https://github.com/mesa/mesa-llm/issues/148

### Changes

- Added `@requires` decorator so tools can declare strict requirements  
  (e.g., `move_one_step` requires an agent position and a model grid).

- Added `tool_filter(tool_name, tool_fn)` method to `LLMAgent`  
  allowing users to dynamically block tools based on agent state  
  (e.g., disable movement when energy is `0`).

- Planners (**CoT**, **ReAct**, **ReWOO**) now use `get_annotated_tools_schema()`  
  - All tools remain visible  
  - Unavailable tools are labeled: `[UNAVAILABLE: reason]`  
  - Helps planners reason about alternatives.

- Executors (tool-calling phase) now use `get_feasible_tools_schema()`  
  - Unavailable tools are removed completely  
  - Prevents the LLM from selecting impossible actions and crashing the simulation.

### Tests

Patch tests added and passing successfully.

test-results

### DIFF
--- a/mesa_llm/__init__.py
+++ b/mesa_llm/__init__.py
@@ -9,6 +9,7 @@ from .parallel_stepping import (
 )
 from .reasoning.reasoning import Observation, Plan
 from .recording.record_model import record_model
+from .tools.tool_decorator import requires
 from .tools.tool_manager import ToolManager
 
 # Enable automatic parallel stepping when mesa_llm is imported
@@ -20,6 +21,7 @@ __all__ = [
     "ToolManager",
     "enable_automatic_parallel_stepping",
     "record_model",
+    "requires",
     "step_agents_parallel",
     "step_agents_parallel_sync",
 ]
--- a/mesa_llm/llm_agent.py
+++ b/mesa_llm/llm_agent.py
@@ -1,3 +1,5 @@
+from collections.abc import Callable
+
 from mesa.agent import Agent
 from mesa.discrete_space import (
     OrthogonalMooreGrid,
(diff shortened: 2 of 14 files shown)


psbuilds:
Hey @hillhack , already working on a fix for this on my local, since I raised the issue first would appreciate if you close this 

Thanks :)

hillhack:
@colinfrisch  can u pls review it

PR #160, opened 2026-03-05
fix: prevent ReWOO replay from mutating cached tool_calls

### LINKED ISSUE #159: ReWOOReasoning silently re-executes the first tool on all replay steps due to in-place mutation of self.current_plan.tool_calls
## Describe the Bug

In `ReWOOReasoning.plan()` and `aplan()`, when replaying a cached multi-step plan (`remaining_tool_calls > 0`), the code aliases `self.current_plan` rather than copying it, then overwrites `.tool_calls` on the alias:

```python
# mesa_llm/reasoning/rewoo.py ~L113–119 (sync); ~L173–179 (async)
index_of_tool = len(self.current_plan.tool_calls) - self.remaining_tool_calls
self.remaining_tool_calls -= 1
tool_call = [self.current_plan.tool_calls[index_of_tool]]
current_plan = self.current_plan          # alias — same object in memory
current_plan.tool_calls = tool_call       # mutates self.current_plan.tool_calls!
```

Because `current_plan` is a reference to the same object, the assignment `current_plan.tool_calls = tool_call` permanently shrinks `self.current_plan.tool_calls` from the full plan list to a single-element list after the very first replay step.

Every subsequent call recalculates the index against a list of length 1, causing Python's negative-index wrapping to silently return the first tool repeatedly:

| Replay step | `len(tool_calls)` | `remaining_tool_calls` | `index` | Tool dispatched |
|-------------|-------------------|------------------------|---------|-----------------|
| 1           | 3                 | 3 -> 2                  | 0       | A ✓             |
| 2           | 1 (mutated)       | 2 -> 1                  | -1      | A ✗             |
| 3           | 1 (mutated)       | 1 -> 0                  | 0       | A ✗             |

For a plan `[A, B, C]`, tools B and C are never executed. No exception is raised — the simulation continues silently with incorrect behaviour.

Both the synchronous `plan()` (~L113–119) and asynchronous `aplan()` (~L173–179) are affected.

## Expected Behavior

Each replay step should dispatch the tool at its correct position in the original plan:

- Step 1 -> A
- Step 2 -> B
- Step 3 -> C

`self.current_plan.tool_calls` must remain unchanged between replay calls so that index calculations stay correct throughout the full replay sequence.

## To Reproduce

1. Configure an agent with `ReWOOReasoning`.
2. Trigger a plan that causes the LLM to return 3 or more tool calls in a single response.
3. Call `agent.step()` (or `agent.astep()`) three times without replanning.
4. Observe that the tool executed in steps 2 and 3 is identical to step 1, regardless of what the LLM originally planned for those positions.

## Additional Context

**Root cause:** `current_plan = self.current_plan` is a reference assignment in Python, not a copy. Mutating `.tool_calls` on the alias writes back to the stored plan.

**Suggested fix:** Take a shallow copy before mutating, so `self.current_plan` is preserved for all remaining replay steps:

```python
import copy

current_plan = copy.copy(self.current_plan)   # shallow copy — does not alias
current_plan.tool_calls = tool_call
return Plan(llm_plan=current_plan, step=self.current_obs.step, ttl=ttl)
```

The same one-line change applies to the async `aplan()` path.

**Regression test:** A test should create a mock plan with tool calls `[A, B, C]`, call `plan()` three times with `remaining_tool_calls` counting down from 3, and assert that the tool name returned in each call is `A`, then `B`, then `C` respectively.

Harsh-617:
Fixes #159

Use copy.copy(self.current_plan) before mutating tool_calls in both plan()
and aplan() so the original plan is preserved across replay steps.

Previously, aliasing self.current_plan caused tool_calls to shrink after
the first replay step, resulting in A->A->A execution instead of A->B->C.

Adds regression tests for both sync and async replay paths.

### DIFF
--- a/mesa_llm/reasoning/rewoo.py
+++ b/mesa_llm/reasoning/rewoo.py
@@ -1,3 +1,4 @@
+import copy
 from typing import TYPE_CHECKING
 
 from mesa_llm.reasoning.reasoning import (
@@ -115,7 +116,7 @@ class ReWOOReasoning(Reasoning):
             )
             self.remaining_tool_calls -= 1
             tool_call = [self.current_plan.tool_calls[index_of_tool]]
-            current_plan = self.current_plan
+            current_plan = copy.copy(self.current_plan)
             current_plan.tool_calls = tool_call
             return Plan(llm_plan=current_plan, step=self.current_obs.step, ttl=ttl)
 
@@ -175,7 +176,7 @@ class ReWOOReasoning(Reasoning):
             )
             self.remaining_tool_calls -= 1
             tool_call = [self.current_plan.tool_calls[index_of_tool]]
-            current_plan = self.current_plan
+            current_plan = copy.copy(self.current_plan)
             current_plan.tool_calls = tool_call
             return Plan(llm_plan=current_plan, step=self.current_obs.step, ttl=ttl)
 
--- a/tests/test_reasoning/test_rewoo.py
+++ b/tests/test_reasoning/test_rewoo.py
@@ -576,6 +576,65 @@ class TestReWOOReasoning:
         assert result3.llm_plan.tool_calls == [mock_tool_3]  # index 2 (3-1=2)
 
 
+    def test_sequential_replay_dispatches_distinct_tools(self):
+        """Regression: plan() replay must dispatch A→B→C, not A→A→A.
+
+        Before the fix, `current_plan = self.current_plan` was an alias, so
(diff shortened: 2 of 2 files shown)


colinfrisch (approved):
I like this and I think that we could merge it soon. I'm going to leave @sanika-n review it as she worked on ReWOO more than me. But don't hesitate if you have any questions

wang-boyu (approved):

PR #167, opened 2026-03-06
fix: make built-in tool injection configurable

### LINKED ISSUE #90: Global Built-in Tool Injection Prevents Strict ToolManager Isolation In Advanced Simulations
Hey @jackiekazil @colinfrisch  @EwoutH ,Hope you are all doing well :)

I’ve been investigating the tool registration behavior in Mesa-LLM and wanted to raise a design/architecture question

### **Summary of the Issue**
The current tool registration system has a **global side-effect**:
built-in tools like `move_one_step`, `teleport_to_location`, and` speak_to` appear to be globally available and appear to be injected directly every time any agent is initialised with tools.

I’m trying to understand whether this is:

- An intentional architectural design choice.
- Or something that might be open to  configurations in the future.

### **Context**
In some models (e.g., spatial social simulations), having movement and communication tools globally available makes sense and is convenient like the example ones in the mesa-llm repo.

But in some models exclusively present inside the examples in the main mesa repo, like the `dining philosophers` example, tools like `move_one_step`, `teleport_to_location` getting injected automatically wouldn't make sense right, it may even lead to hallucinations.

### **Architectural Concern**
I think this is a legitimate gray area,
As simulations become more specialized, this behavior introduces significant friction:

- Capability leakage (e.g. static entities gaining movement)
- Prompt/tool noise for LLMs
- Hard-to-debug conflicts when custom tools shadow built-ins
- Inability to reason about agent permissions locally

I may be misunderstanding the intended design here, so I wanted to check before drawing conclusions
If there is a simple workaround to this please do let me know I probably would have missed it.

Are there plans to make built-in tool injection configurable for more advanced simulations? If so I do have something in mind to tackle this. :)

Would love to understand the intended long-term direction
Thanks :)

yashhzd:
## Summary
Fixes #90

Every `ToolManager` instance automatically copies the global tool registry (`_GLOBAL_TOOL_REGISTRY`), so **all agents** receive `move_one_step`, `teleport_to_location`, and `speak_to` regardless of whether those capabilities make sense for the agent. This causes:

- **Capability leakage** — static entities gain movement tools
- **Tool noise** — LLMs waste reasoning on irrelevant tools
- **Hard-to-debug conflicts** — custom tools can shadow built-ins
- **No local permission reasoning** — cannot restrict agent capabilities

### Changes

**`tool_manager.py`**:
- Add `include_builtins` parameter to `ToolManager.__init__()` (default `True` for full backward compatibility)
- When `False`, the manager starts with an empty tool set — only `extra_tools` (if provided) are registered
- Add `remove_tool(name)` method for fine-grained per-instance tool removal

**`llm_agent.py`**:
- Add `include_builtin_tools` parameter to `LLMAgent.__init__()` that passes through to `ToolManager`

**Tests** (5 new):
- `test_include_builtins_false_starts_empty` — verifies no global tools when disabled
- `test_include_builtins_true_includes_global_tools` — verifies default behavior preserved
- `test_include_builtins_false_with_extra_tools` — verifies only extras are available
- `test_remove_tool` — verifies removal works
- `test_remove_tool_missing_is_silent` — verifies no error on missing tool

### Usage Example

```python
# Agent WITHOUT spatial tools (e.g., dining philosopher)
class Philosopher(LLMAgent):
    def __init__(self, model, **kwargs):
        super().__init__(model, include_builtin_tools=False, **kwargs)
        # Only register tools relevant to this agent
        self.tool_manager.register(pick_up_fork)
        self.tool_manager.register(put_down_fork)

# Agent WITH default tools (backward compatible, no change needed)
class Explorer(LLMAgent):
    def __init__(self, model, **kwargs):
        super().__init__(model, **kwargs)  # gets all built-ins
```

### Test Plan
- [x] All 267 existing tests pass (5 new)
- [x] Pre-commit hooks pass (ruff, codespell)
- [x] Fully backward compatible (default `include_builtins=True`)

### DIFF
--- a/mesa_llm/llm_agent.py
+++ b/mesa_llm/llm_agent.py
@@ -78,7 +78,7 @@ class LLMAgent(Agent):
             api_base=api_base,
         )
 
-        self.tool_manager = ToolManager(include_builtins=include_builtin_tools)
+        self.tool_manager = ToolManager(include_builtin_tools=include_builtin_tools)
         self.vision = vision
         self.reasoning = reasoning(agent=self)
         self.system_prompt = system_prompt
--- a/mesa_llm/llm_agent.py
+++ b/mesa_llm/llm_agent.py
@@ -60,6 +60,7 @@ class LLMAgent(Agent):
         internal_state: list[str] | str | None = None,
         step_prompt: str | None = None,
         api_base: str | None = None,
+        include_builtin_tools: bool = True,
     ):
         super().__init__(model=model)
 
@@ -77,7 +78,7 @@ class LLMAgent(Agent):
             api_base=api_base,
         )
 
-        self.tool_manager = ToolManager()
+        self.tool_manager = ToolManager(include_builtins=include_builtin_tools)
         self.vision = vision
         self.reasoning = reasoning(agent=self)
         self.system_prompt = system_prompt
--- a/mesa_llm/tools/tool_manager.py
+++ b/mesa_llm/tools/tool_manager.py
@@ -45,14 +45,14 @@ class ToolManager:
     def __init__(
         self,
         extra_tools: dict[str, Callable] | None = None,
-        include_builtins: bool = True,
+        include_builtin_tools: bool = True,
     ):
         """
         Args:
             extra_tools: additional tool functions to register on this
-                instance (merged on top of globals when *include_builtins*
+                instance (merged on top of globals when *include_builtin_tools*
                 is True, or used as the sole tool set when False).
-            include_builtins: when True (default) the manager starts with
+            include_builtin_tools: when True (default) the manager starts with
                 every globally-registered tool (move_one_step,
                 teleport_to_location, speak_to, …).  Set to False to
                 create a manager with *only* the tools you explicitly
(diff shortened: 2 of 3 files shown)


mesa_llm/llm_agent.py
```diff
@@ -45,6 +45,7 @@ def __init__(
         vision: float | None = None,
         internal_state: list[str] | str | None = None,
         step_prompt: str | None = None,
+        include_builtin_tools: bool = True,
```
BhoomiAgrawal12: same inconsistent naming as compared to the one used in tool_manager.py

BhoomiAgrawal12 (commented):
hey @yashhzd ! I have added few comments while reviewing the pr, please check it.

yashhzd:
Thanks for the review @BhoomiAgrawal12! All your suggestions were spot on — I've addressed them in the latest push:

1. **Consistent naming** — Renamed `include_builtins` → `include_builtin_tools` everywhere (ToolManager, LLMAgent, tests, docstrings) so the parameter name is uniform across the codebase.

2. **Debug log on `remove_tool`** — Added a `logger.debug()` call when the tool name isn't found, which will surface misspelled names during development without breaking the flow.

3. **Fragile test fix** — Replaced the manual `__tool_schema__` assignment with a proper `@tool` decorator so the test uses the real registration path.

Also rebased on latest `main` to resolve merge conflicts.

PR #181, opened 2026-03-09
add deprecation warnings for pickle-based recordings

apfine:
I propose the following changes to use of `.pkl` in order to ensure the safety of the user's system.

## Summary

This PR keeps pickle support in the recording workflow, but makes its security implications explicit by warning users when pickle-based recordings are loaded or saved.

## Changes

- emit a `UserWarning` when loading `.pkl` recordings in `AgentViewer`
- emit a `UserWarning` when saving recordings with `format="pickle"` in `SimulationRecorder.save()`
- preserve existing pickle support for trusted local workflows
- keep JSON behavior unchanged
- update recording tests to assert warning-based behavior

## Why

Pickle deserialization can execute arbitrary code when loading untrusted files. While pickle may still be useful for trusted local workflows and backward compatibility, it should not appear to be a risk-free default path.

This change keeps user choice while making the security tradeoff explicit at the point of use.

## Validation

- full repository test suite passed
- targeted recording test suite passed
- repo-wide Ruff checks passed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -288,7 +288,6 @@ venv/
 ENV/
 env.bak/
 venv.bak/
-requirements.txt
 
 # Spyder project settings
 .spyderproject
@@ -343,6 +342,3 @@ final_architecture.png
 GSoC_2025.md
 /architectures
 .instructions/
-v
-.github
-
--- a/.gitignore
+++ b/.gitignore
@@ -288,6 +288,7 @@ venv/
 ENV/
 env.bak/
 venv.bak/
+requirements.txt
 
 # Spyder project settings
 .spyderproject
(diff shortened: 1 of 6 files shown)


IlamaranMagesh (commented):
I have reviewed the changes and shared my thoughts in the main conversation thread above. Overall, it looks good. I'll let the maintainers take the call.

wang-boyu:
I updated the pickle load / save warnings to `FutureWarning` and pickle recording support will be removed in the future. Also reverted the unrelated `.gitignore` changes. Merging now.

wang-boyu (approved):

apfine:
> I updated the pickle load / save warnings to `FutureWarning` and pickle recording support will be removed in the future. Also reverted the unrelated `.gitignore` changes. Merging now.

Thankyou so much .

wang-boyu:
Thanks for your work on this @apfine!

PR #185, opened 2026-03-11
fix: ReWOOReasoning.aplan() calls sync add_to_memory instead of async aadd_to_memory

### LINKED ISSUE #184: ReWOOReasoning.aplan() calls sync add_to_memory instead of async aadd_to_memory
**Describe the bug**
While going through the files I found a inconsistency in the way, `add_to_memory` is used for async calls in ReWoo reasoning. It works currently fine as `aadd_to_memory` is also just a wrapper for sync one  in the parent class but in future if anything extra is added to the async parent class then this will violate it and will remain uncaught as that is the purpose this parent class is seperately designed for.

It breaks the async API contract for the memory module. If a Memory subclass overrides aadd_to_memory to do genuinely async work (async lock, async persistence, database write), ReWOOReasoning.aplan() silently bypasses that override and calls the sync base implementation instead.

**Expected behavior**
    await self.agent.memory.aadd_to_memory(   #this should be async
        type="plan",
        content={"content": rsp.choices[0].message.content}
    )

**To Reproduce**
Use ReWOOReasoning with `enable_automatic_parallel_stepping()` or `step_agents_parallel()`. Verify that aplan() is called. The plan is stored via the sync path rather than the async path.

**Additional context**
Both `CoTReasoning.aplan()` and `ReActReasoning.aplan()` correctly use await `aadd_to_memory()` for all memory writes in their async paths and `ReWOOReasoning.aplan()` is the only exception.

This be a small change, but I think should be there.
will be fixing it soon!

BhoomiAgrawal12:
### Pre-PR Checklist
- [T] This PR is a bug fix, not a new feature or enhancement.

### Summary
`ReWOOReasoning.aplan()` was calling the synchronous `add_to_memory()` directly instead of `await self.agent.memory.aadd_to_memory()`.

The `Memory` base class defines both:
- `add_to_memory()` : sync, does in-memory dict manipulation
- `aadd_to_memory()` : async, intended override point for subclasses

If a `Memory` subclass overrides `aadd_to_memory` to do genuinely async work (async database write or async event emit), calling the sync method silently bypasses that override entirely without any error.

All other reasoning classes (`CoTReasoning`, `ReActReasoning`) already use `await aadd_to_memory()` correctly in their async paths. ReWOO was the only one missing it.

### Bug / Issue
Fixes #184 

### Implementation
 [mesa_llm/reasoning/rewoo.py](mesa_llm/reasoning/rewoo.py): replaced `self.agent.memory.add_to_memory(...)` with 
  `await self.agent.memory.aadd_to_memory(...)` in `aplan()`
[tests/test_reasoning/test_rewoo.py](tests/test_reasoning/test_rewoo.py): added `test_aplan_uses_async_memory_method` regression test that asserts `aadd_to_memory` is awaited and `add_to_memory` is never called in the async path


wang-boyu:
Thanks for the fix. I merged `main` branch here to resolve some conflicts. Ready to merge now.

wang-boyu (approved):

PR #188, opened 2026-03-11
feat : Structured llm reasoning added for auditable and constrained decision making

apfine:
## Summary

This PR adds a new structured reasoning strategy, `DecisionReasoning`, on the `cot` branch.

The change does not modify the existing `CoTReasoning` implementation. Instead, it introduces a separate reasoning module for cases where the model should make decisions through a strict, machine-readable schema rather than free-form chain-of-thought text.

The new reasoning path:
- gathers memory and observation context
- asks the LLM for a structured decision object
- validates that object with Pydantic
- stores the structured decision in memory
- executes only the selected `next_action`

### New reasoning module

Added:

- `mesa_llm/reasoning/decision.py`

This module defines:
- `DecisionOption`
- `DecisionOutput`
- `DecisionReasoning`

### New reasoning schema

The model is required to return a strict JSON object containing:

- `goal`
- `constraints`
- `known_facts`
- `unknowns`
- `assumptions`
- `options`
- `chosen_option`
- `rationale`
- `confidence`
- `risks`
- `next_action`

Each option also contains:
- `name`
- `description`
- `tradeoffs`
- `score`

### Validation

The structured response is validated through Pydantic before execution.

This gives:
- a stable output shape
- explicit required fields
- bounded `confidence` in the range `0.0` to `1.0`
- safer downstream use of model output

### Execution flow

The existing tool execution pipeline is preserved.

The flow is now:

1. Build prompt context from memory and current observation
2. Generate a structured decision object
3. Store the decision as memory
4. Extract `next_action`
5. Execute `next_action` through the existing tool executor

This means the executor no longer depends on a long reasoning blob for this reasoning mode.

### Memory integration

The decision artifact is stored as:

- `type="decision"`

This makes the reasoning result easier to inspect and supports future analysis of:
- confidence
- risks
- assumptions
- option selection quality

### Tests

Added:

- `tests/test_reasoning/test_decision.py`

The test coverage includes:
- schema model creation
- prompt generation
- sync planning
- async planning
- selected tool propagation
- prompt fallback behavior
- no-prompt error handling
- execution of `next_action`
- display metadata behavior

## What This PR Does Not Change

This PR does **not** modify:
- `mesa_llm/reasoning/cot.py`
- `CoTReasoning`
- `ReActReasoning`
- `ReWOOReasoning`

So while this branch is named `cot`, the latest change is not an update to the existing CoT implementation. It is a new alternative reasoning mode focused on structured decision-making.

## Why This Change

The current reasoning styles in the repo support:
- free-form chain-of-thought
- lightweight reasoning + action
- multi-step planning

What was missing was a reasoning mode that explicitly separates:
- facts
- unknowns
- assumptions
- options
- tradeoffs
- confidence
- risks
- final executable action

This is useful when decision quality and inspectability matter more than verbose reasoning prose.

## Example Output

```json
{
  "goal": "Reach food",
  "constraints": ["One move this turn"],
  "known_facts": ["Food is visible to the east"],
  "unknowns": ["Whether another agent will block the path"],
  "assumptions": ["The east cell remains traversable this step"],
  "options": [
    {
      "name": "move_east",
      "description": "Move toward visible food",
      "tradeoffs": ["Fast progress", "Potential contention"],
      "score": 0.88
    }
  ],
  "chosen_option": "move_east",
  "rationale": "It best advances the goal with acceptable risk.",
  "confidence": 0.78,
  "risks": ["Another agent may reach the food first"],
  "next_action": "move_east"
}
```

Only `next_action` is forwarded to the executor.

## Validation Performed
The following checks were run:
```python
pytest tests/test_reasoning/test_decision.py tests/test_reasoning -q
ruff check mesa_llm tests
```
Results:

`reasoning tests passed`

Ruff : `passed`

only an existing upstream Mesa deprecation warning was observed during pytest

Files Added : 
`mesa_llm/reasoning/decision.py
tests/test_reasoning/test_decision.py`

## Notes For Reviewers
This PR should be reviewed as:

- a new reasoning capability

- a structured alternative to free-form CoT

- a non-breaking addition to the current reasoning system

### DIFF
--- a/mesa_llm/reasoning/decision.py
+++ b/mesa_llm/reasoning/decision.py
@@ -1,4 +1,3 @@
-import json
 from typing import TYPE_CHECKING
 
 from pydantic import BaseModel, Field
@@ -68,8 +67,18 @@ class DecisionReasoning(Reasoning):
         """
 
     def get_decision_prompt(self, obs: Observation) -> list[str]:
-        prompt_list = [self.agent.memory.get_prompt_ready()]
-        last_communication = self.agent.memory.get_communication_history()
+        prompt_list = []
+
+        get_prompt_ready = getattr(self.agent.memory, "get_prompt_ready", None)
+        if callable(get_prompt_ready):
+            prompt_list.append(get_prompt_ready())
+
+        get_communication_history = getattr(
+            self.agent.memory, "get_communication_history", None
+        )
+        last_communication = (
+            get_communication_history() if callable(get_communication_history) else ""
+        )
 
         if last_communication:
             prompt_list.append("last communication: \n" + str(last_communication))
@@ -77,6 +86,19 @@ class DecisionReasoning(Reasoning):
             prompt_list.append("current observation: \n" + str(obs))
 
         return prompt_list
+
+    def _parse_decision_output(self, response) -> dict:
+        """Normalize structured LLM output into a dictionary."""
+        message = response.choices[0].message
+        parsed = getattr(message, "parsed", None)
+
+        if isinstance(parsed, DecisionOutput):
+            return parsed.model_dump()
+
+        if isinstance(parsed, dict):
+            return DecisionOutput.model_validate(parsed).model_dump()
+
+        return DecisionOutput.model_validate_json(message.content).model_dump()
 
     def plan(
         self,
@@ -112,7 +134,7 @@ class DecisionReasoning(Reasoning):
             response_format=DecisionOutput,
         )
 
-        formatted_response = json.loads(rsp.choices[0].message.content)
+        formatted_response = self._parse_decision_output(rsp)
         self.agent.memory.add_to_memory(type="decision", content=formatted_response)
 
         if hasattr(self.agent, "_step_display_data"):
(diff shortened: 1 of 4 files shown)


mesa_llm/reasoning/decision.py
```diff
@@ -0,0 +1,177 @@
...
+        )
+
+        rsp = await self.agent.llm.agenerate(
+            prompt=prompt_list,
+            tool_schema=selected_tools_schema,
+            tool_choice="none",
+            response_format=DecisionOutput,
+        )
+
+        formatted_response = json.loads(rsp.choices[0].message.content)
```
souro26: `response_format=DecisionOutput` is already passed to the llm call. Parsing json again here ties this code to the raw response format. It might be cleaner to rely on the wrapper's structured output instead.

mesa_llm/reasoning/decision.py
```diff
@@ -0,0 +1,199 @@
...
+
+if TYPE_CHECKING:
+    from mesa_llm.llm_agent import LLMAgent
+
+
+class DecisionOption(BaseModel):
+    name: str
+    description: str
+    tradeoffs: list[str]
+    score: float = Field(
```
souro26: Adding a constraint here would be better in my opinion

souro26:
It's best to wait for a maintainer to review, because there might be things i have missed.

PR #191, opened 2026-03-12
fix(memory): prevent oldest entry from dropping in STLTMemory

### LINKED ISSUE #186: bug: STLTMemory silently drops the oldest memory entry during consolidation
**Describe the bug**

In `STLTMemory._process_step_core()`, when the short-term memory exceeds `capacity + consolidation_capacity`, the oldest entry is removed via `popleft()` *before* `_update_long_term_memory()` is called. Because `_build_consolidation_prompt()` calls `format_short_term()` after the entry has already been evicted, the LLM never sees it. The entry is permanently lost  it never makes it into the long-term summary.

**Expected behavior**

When short term memory overflows, the oldest entry should be summarized into long term memory *before* being evicted. After consolidation, the long-term summary should reflect the content of every entry that was ever held in short-term memory.

**To Reproduce**

```python
from unittest.mock import MagicMock, patch

agent = MagicMock()
agent.step_prompt = None
agent.model.steps = 1

memory = STLTMemory(agent=agent, short_term_capacity=2, consolidation_capacity=1, llm_model="gpt-4o-mini")

# Fill to capacity + consolidation_capacity + 1 to trigger consolidation
for i in range(4):
    memory.step_content = {"observation": f"critical event at step {i}"}
    memory.process_step(pre_step=True)
    agent.model.steps = i + 1
    memory.step_content = {"action": f"response to step {i}"}
    memory.process_step()

# The oldest entry ("critical event at step 0") is never in the consolidation prompt
# Check: long_term_memory does NOT contain "critical event at step 0"
print(memory.long_term_memory)  # step 0 content is silently missing
```

**Additional context**

- Affects `STLTMemory` (the default memory type in `LLMAgent`)
- Affects both `process_step()` (sync) and `aprocess_step()` (async) both delegate to `_process_step_core()` where the premature `popleft()` occurs
- Silent data loss: no error is raised, the agent simply operates with an incomplete long-term summary
- Most impactful in long running simulations where early steps carry critical context (e.g., an agent's initial instructions, opening offers in negotiation, early observations in civil violence models)

---

Harshitmehra-270709:
﻿Fixes #186 

**Describe the bug**
STLTMemory._process_step_core() was previously prematurely popping the oldest entry from the short_term_memory queue before the LLM could summarize it via _update_long_term_memory(). This resulted in the oldest memory being completely lost.

**The Solution**
* Defer the popleft() call to occur exactly *after* the _update_long_term_memory() method calculates the new summary in both process_step and process_step.
* Added a definitive regression test using the repro steps from the issue description to verify the behavior directly against the LLM mock call arguments.

### DIFF
--- a/tests/test_memory/test_STLT_memory.py
+++ b/tests/test_memory/test_STLT_memory.py
@@ -147,9 +147,9 @@ class TestSTLTMemory:
 
         memory._update_long_term_memory()
 
-        assert isinstance(
-            memory.long_term_memory, str
-        ), "long_term_memory must be a string, not a ModelResponse object"
+        assert isinstance(memory.long_term_memory, str), (
+            "long_term_memory must be a string, not a ModelResponse object"
+        )
         assert memory.long_term_memory == "This is the summary text"
 
     def test_observation_tracking(self, mock_agent):
@@ -181,9 +181,9 @@ class TestSTLTMemory:
 
         result = memory.get_prompt_ready()
 
-        assert isinstance(
-            result, str
-        ), f"get_prompt_ready() must return str, got {type(result).__name__}"
+        assert isinstance(result, str), (
+            f"get_prompt_ready() must return str, got {type(result).__name__}"
+        )
         assert "Short term memory:" in result
         assert "Long term memory:" in result
         assert "Test obs" in result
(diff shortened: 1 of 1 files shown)


wang-boyu:
Closing as duplicate of #166.

PR #192, opened 2026-03-12
fix(rewoo): prevent plan() and aplan() replay from mutating cached tool_calls

### LINKED ISSUE #184: ReWOOReasoning.aplan() calls sync add_to_memory instead of async aadd_to_memory

BEASTSHRIRAM:
### Summary

Fixes a silent tool replay bug in `ReWOOReasoning.plan()` and `aplan()` where multi-step tool call replays mutate `self.current_plan.tool_calls` in-place, causing the agent to silently replay incorrect tool steps on subsequent calls. The fix stores the original tool calls separately and returns a copy instead of mutating the cached plan.

### Bug / Issue

Fixes #184 (related to async memory issue, but exposes a broader sync/async mutation bug)

In a complex reasoning task using `ReWOOReasoning` (e.g., the negotiation example 
where an agent plans multiple steps ahead), when a plan has 3+ tool calls:

1. **First replay call:** `remaining_tool_calls=3`, `len(current_plan.tool_calls)=3`, index = 3-3 = 0 ✓ Correct tool returned
2. **First replay mutates:** `current_plan.tool_calls = [tool_0]` (now length 1)
3. **Second replay call:** `remaining_tool_calls=2`, `len(current_plan.tool_calls)=1`, index = 1-2 = -1 ✗ **Negative index wrong tool**
4. **Subsequent calls:** Continue with corrupted indices, agent replays wrong tool step

The mutation happens because both branches of the early-return path used a mutable alias:

```python
# BEFORE (buggy): mutation in both plan() and aplan()
current_plan = self.current_plan
current_plan.tool_calls = tool_call  # ← mutates the cached original
```

No error is raised. The agent silently replays the wrong tool, leading to semantic failures 
in tasks requiring multi-step plans (negotiation, complex reasoning, sequential tool use).

### Implementation

1. Added `self._all_tool_calls = []` to `__init__()` to store the original tool calls list separately
2. When a new plan is generated, store a copy: `self._all_tool_calls = list(rewoo_plan.llm_plan.tool_calls)`
3. Use the original list for index calculation: `index_of_tool = len(self._all_tool_calls) - self.remaining_tool_calls`
4. Return a *copy* of the plan with only the required tool call, leaving the original untouched:
   ```python
   temp_plan = copy.copy(self.current_plan)
   temp_plan.tool_calls = [self.current_plan.tool_calls[index_of_tool]]
   return Plan(llm_plan=temp_plan, ...)
   ```
5. Applied to both `plan()` and `aplan()` methods (identical early-return logic in both)

```python
# _all_tool_calls decouples indexing from mutations
# temp_plan decouples return value from original
# both prevent silent failures on subsequent replays
```

### Testing

All 20 ReWOO reasoning tests pass, including:
- `test_plan_with_remaining_tool_calls()`      verifies correct tool is returned
- `test_aplan_with_remaining_tool_calls()`     async path works correctly  
- `test_remaining_tool_calls_decrement()`     index calculation across multiple scenarios

Tests confirm that each replay returns the correct tool from the full plan without 
corrupting `self.current_plan.tool_calls`.

### Additional Notes

This is a silent correctness bug in `ReWOOReasoning`, the multi-step planning strategy. 
Any agent using ReWOO with plans containing 2+ tool calls is affected. The failure is 
invisible no exception is raised, the agent simply replays wrong tool steps, causing 
task failures that are hard to debug.

No API changes. No new dependencies. The fix adds one `import copy` and decouples state 
management in two methods. Fixes both sync and async paths with identical logic.


wang-boyu:
Thanks for working on this @BEASTSHRIRAM. Could you look into issue #159 and the follow-up fix in PR #160? It appears the replay bug in ReWOO addressed here was already fixed in #160, although with a different approach.

BEASTSHRIRAM:
Hi @wang-boyu,

Thank you for reviewing and for the heads-up about issue #159 / PR #160.

I reviewed PR #160 (Harsh-617's fix). It correctly identifies the aliasing issue and uses `copy.copy(self.current_plan)` before mutation — which is a clean and minimal solution that resolves the immediate bug for both `plan()` and `aplan()`.

However, after digging deeper, I believe there is still a subtle but important gap that this PR addresses differently (and more robustly):

### Key Difference
- **#160** fixes the mutation by copying *just before* assigning the single tool call. This works well.
- **This PR** goes one step further by introducing `_all_tool_calls` as a dedicated, immutable source of truth for indexing + retrieval. It:
  - Stores the *full original* tool calls list when a new plan is generated.
  - Calculates the index from the untouched full list (`len(self._all_tool_calls) - remaining_tool_calls`).
  - Returns a copied plan for the current step.
  - Falls back gracefully in tests (where `current_plan` is sometimes set directly).

This makes the replay logic more resilient, especially in complex scenarios (e.g., 3+ step plans, mixed sync/async usage, or future extensions where the plan object might be modified elsewhere).

### Proposal
I see two good options:

1. **Merge this PR** (or cherry-pick the `_all_tool_calls` approach) — it provides clearer separation of concerns and stronger protection against future similar mutations.
2. **Close this PR** and enhance #160 with the `_all_tool_calls` pattern + the extra regression tests I added.

I’m happy with either  I just want the ReWOO replay behavior to be rock-solid and easy to reason about.

Would you prefer I update this PR to incorporate the best of both, or should we build on #160?

Thanks again!

wang-boyu:
Yes I agree that the approaches are different. However, I do not see any new tests in this PR that would fail on the current `main` branch without this change. As such adding a new `_all_tool_calls` field is just an additional maintenance burden.

If there is no failing reproduction or test against the current `main` branch that #160 does not handle, I would prefer to close this PR as superseded by #160.

BEASTSHRIRAM:
Ok @wang-boyu I understant your concern.You can close this.

PR #208, opened 2026-03-14
Fix: SimulationRecorder.save() appends duplicate simulation_end events on repeated calls

### LINKED ISSUE #201: bug: SimulationRecorder.save() appends duplicate simulation_end events on repeated calls
**Describe the bug**
`SimulationRecorder.save()` unconditionally appends a `simulation_end` event every time it is called. When a user calls `save_recording()` manually and then the `atexit` auto-save handler fires at program exit, the output file contains two `simulation_end` entries.

**Expected behavior**
`simulation_end` is recorded exactly once, regardless of how many times `save()` is called.

**To Reproduce**
```python
import tempfile
from mesa_llm.recording.simulation_recorder import SimulationRecorder

model = MagicMock()
model.steps = 5
model.agents = []
model.__class__.__name__ = "TestModel"
model.max_steps = 10

with tempfile.TemporaryDirectory() as tmp:
    recorder = SimulationRecorder(model=model, output_dir=tmp)
    recorder.save("first_save.json")
    recorder.save("second_save.json")

    ends = [e for e in recorder.events if e.event_type == "simulation_end"]
    print(len(ends))  # this prints 2 where 1 is expected
```

**Additional context**
currently, each call to `save()` appends a new `simulation_end` event.

I will be raising a pr fixing this.

BhoomiAgrawal12:
### Pre-PR Checklist
- [ T] This PR is a bug fix, not a new feature or enhancement.

### Summary
`SimulationRecorder.save()` called `record_model_event( )` unconditionally every time. A manual `save_recording()` followed by the atexit auto-save produced two `simulation_end` entries in the output file.

### Bug / Issue
Fixes #201 

### Implementation
Added a one-line guard:
```python
if not any(e.event_type == "simulation_end" for e in self.events):
```
so `simulation_end` is written at most once per recorder instance.

### Testing
- All existing recording tests pass: `pytest tests/test_recording/ -v`
- Call `recorder.save()` twice and confirmed only one `simulation_end` entry exists

### DIFF
--- a/mesa_llm/recording/simulation_recorder.py
+++ b/mesa_llm/recording/simulation_recorder.py
@@ -241,6 +241,9 @@ class SimulationRecorder:
         )
 
         # Record final model state
+        if any(e.event_type == "simulation_end" for e in self.events):
+            return filepath
+
         self.record_model_event(
             event_type="simulation_end",
             content={
--- a/mesa_llm/recording/simulation_recorder.py
+++ b/mesa_llm/recording/simulation_recorder.py
@@ -134,6 +134,15 @@ class SimulationRecorder:
             else:
                 formatted_content = {"data": content}
 
+        # If recording continues after a save/checkpoint, discard the previous
+        # terminal marker so newly recorded events stay chronologically valid.
+        if (
+            event_type != "simulation_end"
+            and self.events
+            and self.events[-1].event_type == "simulation_end"
+        ):
+            self.events.pop()
+
         # Create the event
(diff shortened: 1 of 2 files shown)


wang-boyu:
Thanks for the fix. I made some changes here to keep `simulation_end` at the end of the event list when `save()` is called multiple times, instead of preserving the first one.

Ready to merge now.

wang-boyu (approved):

PR #210, opened 2026-03-14
Add relevance-based retrieval for episodic memory using cosine similarity

### LINKED ISSUE #209: Implement Relevance Scoring in Episodic Memory
**What's the problem this feature will solve?**
The current EpisodicMemory module successfully implements recency and LLM-graded importance scoring, but is missing the relevance logic that is needed, since the entire memory architecture is inspired from the "Generative Agents" paper, We need to implement vector similarity scoring to complete the "Generative Agents" retrieval architecture.

**Describe the solution you'd like**
Adding a relevancy based scoring logic to the entire file based on a query string which is obtained from the observations stored.

**Are you planning to open a PR for this?**
Yes I'm working on a PR for this issue :)

psbuilds:
Closes #209 

### Summary
As discussed earlier, the EpisodicMemory module successfully utilizes recency and importance scores for retrieval. However, to fully mirror the Generative Agents cognitive architecture (https://github.com/joonspkresearch/generative_agents/blob/fe05a71d3e4ed7d10bf68aa4eda6dd995ec070f4/reverie/backend_server/persona/cognitive_modules/retrieve.py#L175), a third metric is required relevance, computed using a text embedding model and evaluated via cosine similarity.

### Motive
Unlike recency and importance, relevance requires an explicit query string to compare against the stored memory events. This PR introduces the relevance scoring system as a distinct new feature for Episodic Memory, kept entirely separate from general bug fixes and refactoring of the existing scores.

### Relevance Querying 
 Updated retrieve_top_k_entries to accept an explicit `qry_str`.
When an embedding model is provided and a qry_str is passed, the module generates a query embedding and calculates the cosine similarity against the memory entry embeddings. If qry_str is not provided (or an embedding model is not configured), the module falls back to standard recency/importance ranking, preserving backward compatibility.
###  Semantic Embedding Generation & Storage
Introduced self._embeddings (a synchronous deque) to store floating-point representation vectors of memory events alongside self.memory_entries.Created internal _get_embedding and _aget_embedding methods to fetch vector embeddings seamlessly via litellm. The embedding payload is carefully constructed by unpacking the memory 
content dictionary into a clean, natural-language formatted string and raw JSON syntax to maximize the fidelity of the similarity mapping.

### Additional Notes
I did delete some tests which were linked to outdated `process_step` function which was refactored in  #109


ZhehaoZhao423:
Hi @psbuilds, thanks for the openness! I'd be more than happy to push the refactoring for points (1) and (2) to your branch later to help get this moving forward.

Regarding point (3) and the hardcoded return value issue: I completely agree. Relying on a hardcoded magic number (like 0 or 5) when JSON parsing fails will silently skew the agent's memory weighting over time. Instead of a hardcoded fallback, we should ideally leverage a retry mechanism, or at least raise a specific LLMFormatError so the reasoning loop knows the evaluation failed and can handle it contextually.

@wang-boyu, does the core team have a preferred error-handling/fallback pattern for LLM outputs like this moving into Mesa 4.x? Once we have a consensus on the expected behavior, I can push the complete fix for (3) alongside the vectorization updates!

psbuilds:
One small thought for later, since 3 out of the 4 memory classes don't use embeddings, adding embedding directly to the base `MemoryEntry` means they'll all carry an unused field. It might be cleaner down the line to just make a quick subclass (like `EpisodicMemoryEntry`) instead.

Right now, if a user tries to read `entry.embedding` on a basic `ShortTermMemory entry`, it will silently return `None`.

@wang-boyu i think this is ready for a review :)

ZhehaoZhao423:
Great catch, @psbuilds! That’s a very sharp observation and absolutely correct from a strict Object-Oriented Design perspective.

The trade-off I went with here was balancing base-class purity against immediate architectural complexity. Since `embedding` is typed as `list[float] | None = None`, Python handles it as a null reference. This means the actual memory footprint overhead for `ShortTermMemory` and others is practically zero. 
 
If we went the subclass route (`EpisodicMemoryEntry`) right now, we would likely need to refactor the base `Memory` class to support polymorphic entry instantiation (e.g., adding factory methods or overriding the `_finalize_entry` signatures across the board). That felt like it might bloat the scope of this specific PR.
 
That said, I completely agree with you. As the memory modules evolve and require more specialized attributes, subclassing will definitely be the cleaner and more scalable path. Let's absolutely keep your subclassing idea on the radar for a future refactor! Really appreciate the thoughtful review on the PR.

PR #211, opened 2026-03-14
fix: register atexit handler once per class, not once per instance

### LINKED ISSUE #202: bug: `@record_model` registers a new atexit handler on every model instantiation
**Describe the bug**
`atexit.register(_auto_save)` is called inside the `__init__` wrapper created by `@record_model` which means every time a decorated model class is instantiated, a new atexit handler is registered.

In a typical simulation that creates multiple model objects (e.g. for repeated runs, parameter sweeps or tests) this accumulates redundant handlers that all fire at program exit.

**Expected behavior**
At most one atexit handler is registered per recorder instance.

**To Reproduce**
```python
import atexit, tempfile
from mesa.model import Model
from mesa_llm.recording.record_model import record_model

call_count = []
original = atexit.register
atexit.register = lambda f, *a, **k: (call_count.append(f), original(f, *a, **k))

@record_model(output_dir=tempfile.mkdtemp())
class MyModel(Model):
    def __init__(self):
        super().__init__()

MyModel(); MyModel(); MyModel()
print(len(call_count))  # prints 3 where 1 should come
```

**Additional context**
I will be raising a pr for this soon

BhoomiAgrawal12:
### Pre-PR Checklist
- [T ] This PR is a bug fix, not a new feature or enhancement.

### Summary
`@record_model`'s `__init__` wrapper called `atexit.register(_auto_save)` unconditionally on every model instantiation. Creating N model objects registered N atexit handlers, all trying to save the same recording at exit.

### Bug / Issue
Fixes #202 

### Implementation
Added a `_atexit_registered` flag on the recorder instance so the handler is registered exactly once per recorder.
### Additional Notes
- Existing recording tests pass: `pytest tests/test_recording/ -v`
- Created 3 model instances and confirm atexit handler count does not grow

### DIFF
--- a/mesa_llm/recording/record_model.py
+++ b/mesa_llm/recording/record_model.py
@@ -73,6 +73,14 @@ def record_model(
     original_init = cls.__init__
     original_step = getattr(cls, "step", None)
 
+    _save_callbacks: list = []
+
+    def _auto_save_all():
+        for cb in _save_callbacks:
+            cb()
+
+    atexit.register(_auto_save_all)
+
     # ----------------------------- Wrap the model's __init__ method to create and attach the recorder -----------------------------
     @wraps(original_init)
     def init_wrapper(self: "Model", *args, **kwargs):  # type: ignore[override]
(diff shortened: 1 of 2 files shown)


IlamaranMagesh (commented):
> ...all trying to save the same recording at exit.

I don't think so. Even if a model is the same with no change, exp A is different from exp B, so the recordings are different. The recorder also makes sure the files don't get overwritten by creating unique name for each instance.

In my opinion, atexit.register overhead is not a big concern with big-O being still linear. Moreover, this change adds a memory leak which is a major problem and complicates the code maintenance.

I'd like to understand in detail what's the problem this PR addresses.

PR #215, opened 2026-03-14
Feat: Add Long-Term Memory Compression/Pruning to STLTMemory via character bounds

### LINKED ISSUE #214: Feature Request: Memory Decay & Salience Pruning in STLTMemory
#  Feature Request: Memory Decay & Salience Pruning in STLTMemory

## Background & Problem Statement
Currently, [STLTMemory](cci:2://file:///c:/Users/Harshit%20Mehra/OneDrive/Desktop/college/Mesa/mesa-llm/mesa_llm/memory/st_lt_memory.py:9:0-226:9) operates as an append-only summarization pipeline. As the short-term memory exceeds its `capacity` boundaries, older items are summarized and permanently appended to the [long_term_memory](cci:1://file:///c:/Users/Harshit%20Mehra/OneDrive/Desktop/college/Mesa/mesa-llm/mesa_llm/memory/st_lt_memory.py:97:4-105:67) string. 

While this accurately mimics short-to-long term consolidation, the **long-term memory lacks forgetfulness**. Over thousands of simulation steps, the [long_term_memory](cci:1://file:///c:/Users/Harshit%20Mehra/OneDrive/Desktop/college/Mesa/mesa-llm/mesa_llm/memory/st_lt_memory.py:97:4-105:67) string grows indefinitely. This presents two critical problems:
1. **Context Window Limitations**: Prompting the LLM with an ever-expanding summary string will eventually breach context limits or severely increase latency/API costs.
2. **Behavioral Realism**: Real-world entities experience *decay curve memory*—older or less relevant information is naturally forgotten or compressed. Currently, agents possess perfect lifelong recall.

##  Proposed Solution
Introduce a **Memory Decay Mechanism** directly into the STLT architecture.

### Key Additions:
1. **Decay/Salience Factor**: We propose adding a configurable `decay_rate` or a token-limit threshold to [STLTMemory](cci:2://file:///c:/Users/Harshit%20Mehra/OneDrive/Desktop/college/Mesa/mesa-llm/mesa_llm/memory/st_lt_memory.py:9:0-226:9).
2. **Periodic Re-summarization (Pruning)**: Instead of only appending to [long_term_memory](cci:1://file:///c:/Users/Harshit%20Mehra/OneDrive/Desktop/college/Mesa/mesa-llm/mesa_llm/memory/st_lt_memory.py:97:4-105:67), if the long-term string exceeds a specified token limit/character length, the memory system should trigger a "Pruning/Compression" LLM pass.
3. **Prompt Update**:
   ```text
   Your current long-term memory is too long. Please compress it, retaining only the most critical overarching facts and discarding less salient/older details.

Harshitmehra-270709:
## Description
This PR addresses the "Memory Decay & Salience Pruning" proposed feature for `STLTMemory`.

Currently, `STLTMemory` operates such that older items are consolidated and replace the `long_term_memory` string. However, over thousands of steps, the LLM continues to update and include new information. Even with full replacement, this natural tendency of the LLM to retain details causes the long-term string to expand indefinitely, eventually breaching context limits and inflating token processing costs.

This PR introduces a `long_term_char_limit` configuration bounds. If the `long_term_memory` string exceeds this character count after a short-term consolidation, the `STLTMemory` module will trigger a `pruning_prompt` instruction, forcing the LLM to compress the current `long_term_memory` by extracting the overall salient facts and discarding older, less-relevant information. 

## Changes
- Modified `st_lt_memory.py`: Added `long_term_char_limit` to `__init__` and integrated `_prune_long_term_memory` / `_aprune_long_term_memory` triggers after every long-term memory update.
- Updated `STLTMemory` prompts with a new `pruning_prompt` string payload explicitly demanding text-size compression architectures.
- Appended multiple unit tests within `tests/test_memory/test_STLT_memory.py` assessing `long_term_char_limit` threshold breaches and valid secondary LLM consolidation-calls.

## Related Tickets
Closes #214

### DIFF
--- a/mesa_llm/memory/st_lt_memory.py
+++ b/mesa_llm/memory/st_lt_memory.py
@@ -75,7 +75,7 @@ class STLTMemory(Memory):
             You are a helpful assistant that compresses excessively long text memories.
             The following text represents an agent's long-term memory. It has grown too large.
             Please compress it, retaining only the most critical overarching facts and lessons, and discarding less salient or outdated details.
-            
+
             Current Long-term memory to prune:
             {long_term_memory}
             """
--- a/tests/test_memory/test_STLT_memory.py
+++ b/tests/test_memory/test_STLT_memory.py
@@ -147,9 +147,9 @@ class TestSTLTMemory:
 
         memory._update_long_term_memory()
 
-        assert isinstance(
-            memory.long_term_memory, str
-        ), "long_term_memory must be a string, not a ModelResponse object"
+        assert isinstance(memory.long_term_memory, str), (
+            "long_term_memory must be a string, not a ModelResponse object"
+        )
         assert memory.long_term_memory == "This is the summary text"
 
     def test_observation_tracking(self, mock_agent):
@@ -181,9 +181,9 @@ class TestSTLTMemory:
 
         result = memory.get_prompt_ready()
 
-        assert isinstance(
-            result, str
-        ), f"get_prompt_ready() must return str, got {type(result).__name__}"
+        assert isinstance(result, str), (
+            f"get_prompt_ready() must return str, got {type(result).__name__}"
+        )
         assert "Short term memory:" in result
         assert "Long term memory:" in result
         assert "Test obs" in result
@@ -195,9 +195,9 @@ class TestSTLTMemory:
 
         result = memory.get_prompt_ready()
 
-        assert isinstance(
-            result, str
-        ), f"get_prompt_ready() must return str, got {type(result).__name__}"
+        assert isinstance(result, str), (
+            f"get_prompt_ready() must return str, got {type(result).__name__}"
+        )
         assert "Short term memory:" in result
         assert "Long term memory:" in result
 
@@ -233,9 +233,9 @@ class TestSTLTMemory:
         call_args = mock_llm.generate.call_args[0][0]
 
         # The oldest entry MUST be in the prompt sent to the LLM
-        assert (
-            "critical event at step 0" in call_args
-        ), "Oldest memory entry was silently dropped before consolidation!"
+        assert "critical event at step 0" in call_args, (
+            "Oldest memory entry was silently dropped before consolidation!"
+        )
 
     def test_memory_pops_and_summarizes_exact_k_items(
         self, mock_agent, mock_llm, llm_response_factory
@@ -339,7 +339,6 @@ class TestSTLTMemory:
         """
         Verify that _aprune_long_term_memory behaves the same asynchronously
         """
-        import asyncio
 
         async def mock_agenerate(prompt):
             if "compresses excessively long text memories" in prompt:


IlamaranMagesh (commented):
**Context**
[my comment on the issue](https://github.com/mesa/mesa-llm/issues/214#issuecomment-4061043510)

I get it. My opinions are,

Long term memory can't go out of context, unless user switch models in between with different context windows

> ...string over time to retain all historical details

Isn't this what we want? I get that this doesn't mimic necessarily the exact long-term memory and moreover, information is already compressed when the long-term memory is consolidated but I'm open to discuss with others.

Harshitmehra-270709:
> **Context** [my comment on the issue](https://github.com/mesa/mesa-llm/issues/214#issuecomment-4061043510)
> 
> I get it. My opinions are,
> 
> Long term memory can't go out of context, unless user switch models in between with different context windows
> 
> > ...string over time to retain all historical details
> 
> Isn't this what we want? I get that this doesn't mimic necessarily the exact long-term memory and moreover, information is already compressed when the long-term memory is consolidated but I'm open to discuss with others.

That makes total sense! My thought process was geared toward users running very long simulations locally using smaller open-source models (like Llama 3 8B or Mistral) where context windows are much smaller (8k natively) and latency is a bigger concern. Also, in some cognitive architectures, having an optional 'decay' factor helps agents adapt solely to recent paradigms rather than being overly biased by ancient history!

However, I completely agree that if a user is running a modern mega-context model, retaining all historical details is strictly preferred.

I really appreciate you taking the time to review it and outline your thoughts! I'm happy to leave this open for discussion with the rest of the community. If others decide they'd find this useful as an optional config later on, the PR is ready to go. Thanks again!

IlamaranMagesh:
> smaller open-source models (like Llama 3 8B or Mistral) where context windows are much smaller

Here the model size doesn't matter. Like I said before, if an agent runs with the same model throughout the simulation, it doesn't go out of context. It's because, the long-term memory is consolidated by the model itself so, the output will always be within the model context.

The question here is, 'do we need a system to filter out less relevant long-term memory?'. My answer would be, 'yes, but not now' as this is a feature on top of a feature. There are other places where the efforts are needed the most now. But I agree with you to have the discussion open.

Harshitmehra-270709:
> > smaller open-source models (like Llama 3 8B or Mistral) where context windows are much smaller
> 
> Here the model size doesn't matter. Like I said before, if an agent runs with the same model throughout the simulation, it doesn't go out of context. It's because, the long-term memory is consolidated by the model itself so, the output will always be within the model context.
> 
> The question here is, 'do we need a system to filter out less relevant long-term memory?'. My answer would be, 'yes, but not now' as this is a feature on top of a feature. There are other places where the efforts are needed the most now. But I agree with you to have the discussion open.

Totally fair point! The self-constraining nature of LLM-generated summaries is a really elegant property I hadn't fully considered. I'll look into the other open issues and features that can be implemented!

Do you have any recommendations about the features that can be added?
Can you mail them to me? My mail ID is Mehra.harshit27@gmail.com.

Thanks for the thoughtful discussion!

PR #221, opened 2026-03-14
fix: add resilient parallel execution with per-agent error isolation

### LINKED ISSUE #220: bug: Parallel stepping crashes entire simulation when a single agent's step fails
## Bug Description

When using parallel agent stepping (`step_agents_parallel`, `step_agents_multithreaded`, or `do_async`), if any single agent's `astep()` or `step()` raises an exception, ALL other agent tasks are cancelled/abandoned and the entire simulation crashes.

This is particularly problematic for LLM-backed agents where failures are common (rate limits, timeouts, malformed JSON responses, etc.).

## Root Cause

Three parallel execution paths lack error isolation:

1. `step_agents_parallel()` — `asyncio.gather(*tasks)` without `return_exceptions=True`
2. `step_agents_multithreaded()` — `future.result()` loop without try/except
3. `_agentset_do_async()` — same `asyncio.gather` issue

## Reproduction

```python
class FailingAgent(LLMAgent):
    async def astep(self):
        raise RuntimeError("LLM timeout")

class WorkingAgent(LLMAgent):
    async def astep(self):
        self.counter += 1

# Both agents crash — WorkingAgent never completes
await step_agents_parallel([FailingAgent(m), WorkingAgent(m)])
```

## Expected Behavior

- Failed agents are isolated: other agents complete normally
- Failures are logged with agent ID and exception details
- Users can inspect which agents failed and why
- Optional "raise" mode for debugging

yashhzd:
## Summary

Closes #220

Fixes a critical bug where one agent's exception during parallel stepping crashes ALL other agent tasks. This is devastating for LLM-backed simulations where failures (rate limits, timeouts, JSON parse errors) are expected and common.

### `AgentStepResult` dataclass
- Structured result with `agent`, `success`, and `exception` fields
- Returned by all parallel stepping functions

### Error isolation in all 3 parallel paths
- `step_agents_parallel()` — `asyncio.gather(return_exceptions=True)` + per-result error logging
- `step_agents_multithreaded()` — try/except around `future.result()` + error logging
- `_agentset_do_async()` — same `return_exceptions=True` treatment

### Configurable error handling
- `on_error="continue"` (default) — log and continue, fully backward compatible
- `on_error="raise"` — propagate first exception for debugging
- Optional per-agent `timeout` — prevents one slow LLM call from blocking everything
- Config via function params or `enable_automatic_parallel_stepping(on_error=..., timeout=...)`

## Backward Compatibility

- Default `on_error="continue"` is strictly better than current behavior (crashes → logs & continues)
- All existing function signatures unchanged (new params are keyword-only with defaults)
- All 4 existing `test_parallel_stepping.py` tests pass unchanged
- Return type changes from `None` to `list[AgentStepResult]` — no existing code inspects the return

## Test Plan

- [x] All existing parallel stepping tests pass (4/4)
- [x] 13 new tests covering: error isolation (async + threaded), raise mode, timeout, do_async resilience, backward compat, config validation
- [x] Ruff lint and format checks pass

### DIFF
--- a/mesa_llm/__init__.py
+++ b/mesa_llm/__init__.py
@@ -3,6 +3,7 @@ import datetime
 import mesa_llm.tools.inbuilt_tools  # noqa: F401, to register inbuilt tools
 
 from .parallel_stepping import (
+    AgentStepResult,
     enable_automatic_parallel_stepping,
     step_agents_parallel,
     step_agents_parallel_sync,
@@ -15,6 +16,7 @@ from .tools.tool_manager import ToolManager
 enable_automatic_parallel_stepping()
 
 __all__ = [
+    "AgentStepResult",
     "Observation",
     "Plan",
     "ToolManager",
--- a/mesa_llm/parallel_stepping.py
+++ b/mesa_llm/parallel_stepping.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 import asyncio
 import concurrent.futures
 import logging
+from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
 from mesa.agent import Agent, AgentSet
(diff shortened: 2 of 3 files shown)


IlamaranMagesh (commented):
I see that you've tried to solve the AI model failure problem. The solution is right in most of the cases but here the solution silently corrupts the simulation.

ABM is about agents. If one agent is stale and acts on data that is old, the simulation is wrong. Here the failed agent is neither falling back to a default behaviour nor it's retrying the step. The solution silenty moves to the next step. Consider this happening for the same agent N times, now the agent is behind N steps and other agents acted on this stale agent. Such scenarios make the framework unreliabie and unstable.

This needs a detailed discussion as it's more of a feature than a bug.

yashhzd (commented):
Thanks for the thoughtful review — this is a valid concern and worth discussing.

You're right that a silently stale agent can corrupt downstream simulation state. But I think it's important to weigh that against the **current behavior**, which is strictly worse: one agent's failure crashes the entire simulation, killing all 99 healthy agents along with it. There's no recovery path at all.

This PR moves from "everyone dies" to "the failed agent is inert but everyone else continues." That's not perfect, but it's a meaningful improvement — especially for LLM-backed simulations where transient failures (rate limits, timeouts, malformed JSON) are expected and often resolve on the next step.

That said, I agree that "silently continue" shouldn't be the only option. A few ideas for extending this:

1. **`on_failure` callback** — users could register a handler like `on_failure=lambda agent, exc: agent.fallback_step()` to define custom recovery (retry, remove agent, use a default action, etc.)
2. **Automatic retry** — a `max_retries` parameter on the parallel stepper that retries failed agents before giving up
3. **Agent removal** — an `on_error="remove"` mode that removes the failed agent from the simulation entirely

I intentionally kept retry logic out of this PR because `ModuleLLM` already has tenacity-based retry at the LLM call level. If an agent still fails after those retries, the failure is likely not transient. But a higher-level retry or callback at the stepping layer could complement that.

Happy to add an `on_failure` callback hook in a follow-up if the maintainers think that's the right direction. The current PR at least gives users `AgentStepResult` to inspect failures and act on them in their model's `step()` method — which is already more than what exists today (an unhandled crash).

PR #224, opened 2026-03-14
fix: prevent double pre_step/post_step when astep() delegates to wrapped step()

### LINKED ISSUE #222: bug: Double pre_step/post_step calls when astep() delegates to wrapped step() causes orphaned memory entries
## Bug Description

When a subclass of `LLMAgent` defines only `step()` (no `astep()`), the default `astep()` method calls `apre_step()` and `apost_step()` — but then delegates to `self.step()`, which is **already wrapped** by `__init_subclass__` with its own `pre_step()` and `post_step()`. This causes `process_step(pre_step=True)` and `process_step(pre_step=False)` to each be called **twice per step** during parallel execution.

For `STLTMemory` (the default memory), this creates an **orphaned `MemoryEntry`** with `step=None` that is never finalized or cleaned up.

## Environment

- **mesa-llm**: v0.3.0 (commit `a8161c7`)
- **mesa**: 3.5.0
- **Python**: 3.12

## Root Cause

In [`llm_agent.py`](https://github.com/mesa/mesa-llm/blob/a8161c735478bc7aeb4911335954029de53b3728/mesa_llm/llm_agent.py), the default `astep()` (line 318) and `__init_subclass__` wrapping (line 331) both inject pre/post step hooks:

**Default `astep()` (line 318–329):**
```python
async def astep(self):
    await self.apre_step()          # ← First pre_step call
    if hasattr(self, "step") and self.__class__.step != LLMAgent.step:
        self.step()                 # ← This is the WRAPPED step
    await self.apost_step()         # ← Second post_step call
```

**`__init_subclass__` wrapping (line 340–351):**
```python
if user_step:
    def wrapped(self, *args, **kwargs):
        LLMAgent.pre_step(self)     # ← Second pre_step call
        result = user_step(self)
        LLMAgent.post_step(self)    # ← First post_step call
        return result
    cls.step = wrapped
```

When `astep()` calls the wrapped `self.step()`, the actual call sequence is:

| # | Call | Effect on STLTMemory |
|---|------|---------------------|
| 1 | `apre_step()` | Creates **entry1** (with step_content data, step=None), resets `step_content={}` |
| 2 | wrapped → `pre_step()` | Creates **entry2** (empty content, step=None) |
| 3 | wrapped → `user_step()` | User's logic runs, may add to step_content |
| 4 | wrapped → `post_step()` | Pops **entry2** (last), merges with step_content, creates **real_entry** |
| 5 | `apost_step()` | Sees last entry has `step != None` → early return, no-op |

**Result:** entry1 is permanently orphaned in `short_term_memory` with `step=None`.

## Impact

- **Memory pollution**: Orphaned entries with `step=None` accumulate in short-term memory (one per step during parallel execution)
- **Premature consolidation**: Orphaned entries count toward `len(short_term_memory)`, triggering early `popleft()` eviction of real entries
- **LLM context confusion**: `format_short_term()` renders orphaned entries as `"Step None: {data}"`, polluting the prompt sent to the LLM
- **Data loss**: The first `process_step(pre_step=True)` captures the original `step_content` but that entry is never finalized — the actual data ends up orphaned while the second (empty) entry gets processed

## Reproduction

```python
import asyncio
from mesa.model import Model
from mesa_llm.reasoning.cot import CoTReasoning

class DummyModel(Model):
    def __init__(self):
        super().__init__(seed=42)

class MyAgent(LLMAgent):
    def __init__(self, model):
        super().__init__(model, reasoning=CoTReasoning, llm_model="gemini/gemini-2.0-flash")

    def step(self):  # Only defines step(), no astep()
        pass

async def main():
    m = DummyModel()
    agent = MyAgent(m)

    # Simulate what parallel stepping does
    await agent.astep()

    # Check for orphaned entries
    orphaned = [e for e in agent.memory.short_term_memory if e.step is None]
    print(f"Orphaned entries: {len(orphaned)}")  # Expected: 0, Actual: 1

    # After 10 steps, there are 10 orphaned entries
    for _ in range(9):
        await agent.astep()
    orphaned = [e for e in agent.memory.short_term_memory if e.step is None]
    print(f"Orphaned entries after 10 steps: {len(orphaned)}")  # 10!

asyncio.run(main())
```

## Who Is Affected

Any `LLMAgent` subclass that:
1. Defines only `step()` (no custom `astep()`) — the most common pattern
2. Is executed via `astep()` during parallel stepping — the default execution mode

This means the bug affects **most mesa-llm users** running parallel simulations with the default `STLTMemory`.

## Expected Behavior

- `pre_step` and `post_step` should be called **exactly once** per step, regardless of whether execution goes through `step()` or `astep()`
- No orphaned `MemoryEntry` objects with `step=None` should persist in memory

yashhzd:
## Summary

Closes #222

Fixes a bug where `pre_step`/`post_step` (and their async variants) are called **twice per step** when a subclass defines only `step()` and is executed via `astep()` during parallel stepping. This is the most common usage pattern in mesa-llm.

### Root Cause

The default `LLMAgent.astep()` calls `apre_step()` → `self.step()` → `apost_step()`. But `self.step()` is the **wrapped** version from `__init_subclass__`, which also calls `pre_step()` → `user_step()` → `post_step()`. Result: memory hooks fire twice.

### The Fix

Store the raw user `step()` function as `cls._raw_user_step` during `__init_subclass__` wrapping. The default `astep()` calls `_raw_user_step` directly (bypassing the pre/post wrapper), since it handles its own async pre/post hooks.

| Path | Before | After |
|------|--------|-------|
| `astep()` → step-only subclass | pre_step ×2, post_step ×2 | pre_step ×1, post_step ×1 |
| `astep()` → astep-only subclass | pre_step ×1 (already correct) | unchanged |
| `step()` called directly | pre_step ×1 (already correct) | unchanged |

### Impact of the Bug

- **Memory pollution**: Orphaned `MemoryEntry` with `step=None` accumulates in STLTMemory (one per step)
- **Premature consolidation**: Orphaned entries count toward memory capacity, evicting real entries early
- **LLM confusion**: `format_short_term()` renders orphaned entries as `"Step None: {data}"` in the prompt

## Backward Compatibility

- All 281 existing tests pass unchanged
- Direct `step()` calls still go through the wrapper (no change)
- Only the `astep()` → `step()` delegation path is fixed
- `_raw_user_step` is a private class attribute, no public API change

## Test Plan

- [x] All 281 existing tests pass
- [x] 6 new tests covering:
  - `astep()` with step-only agent: exactly 1 pre/post call
  - `astep()` with async-only agent: exactly 1 pre/post call
  - `astep()` with both-methods agent: exactly 1 pre/post call
  - Direct `step()` call: exactly 1 pre/post call
  - 5 consecutive `astep()` calls: counts == 5, not 10
  - Real STLTMemory: no orphaned entries with `step=None`
- [x] Ruff lint and format clean


wang-boyu:
Thanks for the fix. I have merged `main` to the PR branch to resolve conflicts introduced by another recently merged PR.

Merging now.

wang-boyu (approved):

PR #225, opened 2026-03-15
fix: selected_tools=[] should return no tools, not all tools

AdityaChauhanX07:
`get_all_tools_schema()` checks `if selected_tools:` to decide whether to filter tools.
The problem is that `[]` is falsy in Python, so `selected_tools=[]` hits the `else` branch and returns every registered tool - the opposite of what the caller asked for.

I ran into this while building models in my [GSoC learning space](https://github.com/AdityaChauhanX07/GSoC-learning-space).
My `hello_llm_agent` model passes `selected_tools=[]` with a system prompt saying "never use tools," but agents were still calling `teleport_to_location` because the schemas were getting injected anyway.

The fix is changing `if selected_tools:` to `if selected_tools is not None:` so that:
- `None` → all tools (default, unchanged)
- `[]` → no tools (the actual opt-out)
- `["tool_a"]` → just those tools (unchanged)

Also updated `test_get_all_tools_schema_empty_list` which was asserting the buggy behavior as "current behavior."

All 275 tests pass, 0 failures.

**Note:** This is technically a breaking change for any code that passed `selected_tools=[]` expecting all tools. The previous behavior was a bug - the test documenting it was labeled "current behavior" not "intended behavior."

### DIFF
--- a/mesa_llm/reasoning/cot.py
+++ b/mesa_llm/reasoning/cot.py
@@ -93,7 +93,12 @@ class CoTReasoning(Reasoning):
         selected_tools: list[str] | None = None,
     ) -> Plan:
         """
-        Plan the next (CoT) action based on the current observation and the agent's memory.
+        Plan the next (CoT) action based on the current observation and memory.
+
+        ``selected_tools`` is forwarded to ``ToolManager.get_all_tools_schema()``.
+        Omitting it or passing ``None`` uses the default behavior of exposing
+        all tools, ``[]`` exposes no tools, and a non-empty list restricts
+        planning/execution to the named tools.
         """
         # If no prompt is provided, use the agent's default step prompt
         if prompt is None:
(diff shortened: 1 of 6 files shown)


AdityaChauhanX07:
Thanks for the review. Good point on the breaking change note - though I'd argue any caller passing `selected_tools=[]` and expecting all tools back was relying on buggy behavior, not intentional API design. The test itself was documented as "current behavior" not "intended behavior." But worth mentioning, so I'll add a note to the PR description.

wang-boyu:
Thanks for the fix. I removed `Closes #148` from the PR description, so that #148 won't be closed automatically.

The broader question in #148 about feasible tool exposure still needs more design discussion, and PR #168 is already exploring that direction. I'd prefer to keep this PR scoped to the empty-list fix and leave #148 open for the larger follow-up.

Also updated some docstrings to document how `selected_tools` behaves. Merging now.

PR #231, opened 2026-03-16
fix: prevent JSON serialization errors in send_message/asend_message …

### LINKED ISSUE #156: send_message/asend_message store Agent objects in memory content causing JSON serialization errors
**Describe the bug**

`LLMAgent.send_message` and `LLMAgent.asend_message` store `sender` and `recipients`
as Python Agent objects inside the memory `content` dict. Because Agent objects are not
JSON serializable, this silently corrupts memory content and causes a hard crash whenever
recorded events are saved to JSON (e.g. via `SimulationRecorder`).

The built-in `speak_to` tool already stores `unique_id` values instead of objects, so
the two messaging mechanisms are currently inconsistent.

**To Reproduce**

```python
# Crash via SimulationRecorder
@record_model
class MyModel(mesa.Model):
    ...

model = MyModel()
agent_a.send_message("hello", [agent_b])
model.save_recording()          # TypeError: Object of type LLMAgent is not JSON serializable

# Or directly:
import json
agent_a.send_message("hello", [agent_b])
json.dumps(agent_a.memory.step_content)  # same error

```

**Expected behavior**

Message memory entries should store serializable identifiers:
```
content={
    "message": message,
    "sender": self.unique_id,
    "recipients": [r.unique_id for r in recipients],
}
```
This is already what speak_to does (llm_agent.py lines 213–218) and what
serialization-safe memory handling requires.

**Current behavior**

Both send_message (line 284) and asend_message (line 269) in llm_agent.py store:
```
content={
    "message": message,
    "sender": self,           # Agent object — not serializable
    "recipients": recipients, # list of Agent objects — not serializable
}
```
**Additional context**

Affects: llm_agent.py lines 269–273 (asend_message) and 284–288 (send_message)
The bug is silent until JSON serialization is attempted, making it hard to catch without recording enabled
speak_to (inbuilt_tools.py lines 213–218) handles this correctly and should be the reference implementation

adarshkumar23:
Fixes #156

## Problem
`send_message` and `asend_message` store Agent objects in memory 
content and return strings, causing JSON serialization errors when 
memory is serialized for LLM context.

Additionally, no safety check exists for recipients without a 
`.memory` attribute (non-LLM agents).

## Solution
- Compute `recipient_ids` as a list of `unique_id` integers 
  (JSON-serializable) instead of referencing Agent objects
- Use `getattr(recipient, "memory", None)` to safely handle 
  non-LLM agent recipients
- Return string uses `self.unique_id` and `recipient_ids` 
  instead of Agent object representations

## Related
- Similar pattern to PR #229 which fixed the same issue in `speak_to`
image


IlamaranMagesh (commented):
The issue is closed and fix #157 to the issue was already merged. Could you please provide details what's the PR is trying to solve?

mesa_llm/llm_agent.py
```diff
@@ -262,33 +262,39 @@ async def asend_message(self, message: str, recipients: list[Agent]) -> str:
...
                 content={
                     "message": message,
                     "sender": self.unique_id,
-                    "recipients": [r.unique_id for r in recipients],
+                    "recipients": recipient_ids,
                 },
             )
-
-        return f"{self} → {recipients} : {message}"
+        return f"{self.unique_id} → {recipient_ids} : {message}"
```
IlamaranMagesh: Fix 2: `recipients` was still list[Agents], though it is not parsed as JSON and used to return a String back, adds clarity since there is not `__str__` or similar implementation in `mesa.Agent` base class.

IlamaranMagesh (commented):
I've added some clarity to the fixes on what this PR is trying to do.

For context, there is a concern raised on adding `__str__` to Base classes [here](https://github.com/mesa/mesa/issues/3625)

wang-boyu:
Thanks for the PR. Part of it duplicated what was already fixed in #157, but it also checks non-LLM recipients without memory (what was done to `speak_to` in #161) and updates the return string.

I have merged `main` branch into this branch to resolve conflicts, and updated `send_message` / `asend_message` to align their behavior with `speak_to` (see https://github.com/mesa/mesa-llm/pull/161#issuecomment-4061955886).

LGTM now. Merging.

wang-boyu (approved):

PR #232, opened 2026-03-16
fix(llm_agent): initialize self.recorder so @record_model can attach it

### LINKED ISSUE #218: bug: self.recorder not initialised in LLMAgent
## Describe the bug

The recorder attachment logic in `record_model.py` uses `hasattr(agent, "recorder")` to determine whether to attach the simulation recorder to agents. However, the LLMAgent has no `recorder` attribute defined at class level, so `hasattr` returns `False` and the recorder is never actually attached to agents, unless the user creates it in the subclass.

```python

# record_model.py
def _attach_recorder_to_agents(model, recorder):
    for agent in list(model.agents):
        if hasattr(agent, "recorder"):   # ← always False for LLMAgent
            agent.recorder = recorder

```

## Expected behavior

LLMAgent class should have `self.recorder: SimulationRecorder | None = None` attribute set

## Am I going to open a PR?

Yes, I'll be opening it after I get some clarification on #219

## Additional context

Even with this bug resolved, the agents' events are not recorded because the `record_event` method is not called by the agents. (here, #219)

yashhzd:
## Summary

`LLMAgent.__init__` never initialized `self.recorder`, so the `hasattr(agent, "recorder")` check in `_attach_recorder_to_agents()` always returned `False`. This meant the `@record_model` decorator silently never attached the `SimulationRecorder` to any `LLMAgent` instance.

### Changes

- **`mesa_llm/llm_agent.py`**: Add `self.recorder: SimulationRecorder | None = None` in `__init__`, with a `TYPE_CHECKING` import to avoid circular imports.
- **`tests/test_llm_agent.py`**: Add unit test verifying `hasattr(agent, "recorder")` returns `True` and defaults to `None`.
- **`tests/test_recording/test_record_model.py`**: Add integration test verifying `@record_model` successfully propagates the recorder to `LLMAgent` instances created during model init.

## Test plan

- [x] `pytest tests/test_llm_agent.py::test_llm_agent_has_recorder_attribute -xvs` passes
- [x] `pytest tests/test_recording/test_record_model.py::TestRecordModelDecorator::test_recorder_attached_to_llm_agents -xvs` passes
- [x] Full test suite (284 tests) passes with no regressions
- [x] `ruff check` and all pre-commit hooks pass

Closes #218

### DIFF
--- a/mesa_llm/llm_agent.py
+++ b/mesa_llm/llm_agent.py
@@ -1,5 +1,8 @@
+from __future__ import annotations
+
 import logging
 import warnings
+from typing import TYPE_CHECKING
 
 from mesa.agent import Agent
 from mesa.discrete_space import (
@@ -23,6 +26,9 @@ from mesa_llm.reasoning.reasoning import (
 from mesa_llm.tools.tool_manager import ToolManager
 
 logger = logging.getLogger(__name__)
+
+if TYPE_CHECKING:  # pragma: no cover
+    from mesa_llm.recording.simulation_recorder import SimulationRecorder
 
 
 class LLMAgent(Agent):
@@ -85,6 +91,9 @@ class LLMAgent(Agent):
 
         # display coordination
         self._step_display_data = {}
+
+        # Placeholder so @record_model can attach the SimulationRecorder
+        self.recorder: SimulationRecorder | None = None
 
         if isinstance(internal_state, str):
             internal_state = [internal_state]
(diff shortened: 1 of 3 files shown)


IlamaranMagesh (commented):
Thanks for opening a PR. Though this is clearly a bug, I'm still waiting for some clarifications from the maintainers (#219) that I mentioned in the issue #218

wang-boyu:
LGTM. Merging. Thanks for the issue, the fix, and the discussions.

wang-boyu (approved):

PR #233, opened 2026-03-16
warn when LLMAgent.astep() falls back to blocking step()

### LINKED ISSUE #140: bug: LLMAgent.astep() blocks the event loop when subclass only defines step()
**Describe the bug**
When a user subclasses `LLMAgent` and only defines `step()` without `astep()`, calling `astep()` falls back to invoking `step()` synchronously inside an async method. Since `step()` makes blocking LLM API calls, this freezes the entire event loop and prevents other agents from running concurrently.

**Expected behavior**
`astep()` should remain non-blocking even when falling back to `step()`, so all agents in a parallel simulation can run concurrently without freezing the event loop.

**To Reproduce**
Subclass `LLMAgent` with only a `step()` method and run a parallel simulation. Despite the async setup, agents execute one at a time due to the blocking fallback in `astep()`:

```python
async def astep(self):
    await self.apre_step()
    if hasattr(self, "step") and self.__class__.step != LLMAgent.step:
        self.step()   # ← blocks the event loop
    await self.apost_step()
```

**Additional context**
Fix: offload the blocking `step()` call to a thread pool executor so the event loop stays free:

```python
loop = asyncio.get_event_loop()
await loop.run_in_executor(None, self.step)  # non-blocking
```

adarshkumar23:
Fixes #140

## Problem
When a subclass only defines `step()` (not `astep()`), the default 
`astep()` calls `self.step()` synchronously inside the async method. 
This blocks the entire event loop, preventing parallel agent execution.

Additionally, `astep()` was calling `apre_step()`/`apost_step()` before 
delegating to the wrapped `step()`, which already calls 
`pre_step()`/`post_step()` internally — causing double execution of 
step hooks.

## Solution
- Use `asyncio.to_thread(self.step)` to run the sync fallback in a 
  thread pool, keeping the event loop free for other agents
- Skip async pre/post step hooks when delegating to wrapped `step()` 
  since it already handles its own pre/post hooks via `__init_subclass__`
- Keep `apre_step()`/`apost_step()` for the else branch (when no 
  custom step exists)

## Testing
- All 282 existing tests pass
Screenshot 2026-03-16 213606
Screenshot 2026-03-16 213638


wang-boyu:
As noted in https://github.com/mesa/mesa-llm/issues/140#issuecomment-4063406604, #140 is not a bug but intended behavior. I therefore reverted the original behavior changes in this PR and replaced them with a warning when `astep()` falls back to synchronous `step()`, so users are reminded to override `astep()` if they need non-blocking behavior. I also merged `main` to resolve conflicts.

Merging now.

wang-boyu (approved):

PR #165, opened 2026-03-06
fix: accumulate additive events in memory staging area

yashhzd:
## Summary
Fixes #137

The base `Memory.add_to_memory()` method uses a plain dict (`step_content`) where each event type maps to a single value. When multiple events of the same type occur within a single step (e.g., an agent receiving messages from several neighbors), subsequent calls silently overwrite previous entries, causing **permanent data loss**.

### Changes

- **`memory.py`**: Introduce `ADDITIVE_EVENT_TYPES` (`frozenset({"message", "action"})`) — events of these types are collected in a list rather than overwritten. Observations remain state-based (overwrite semantics preserved). Other types (plans, reasoning steps, etc.) also keep overwrite semantics.
- **`memory.py` (`MemoryEntry.__str__`)**: Updated to render list-valued content with numbered sub-entries for clean display.
- **`st_memory.py` / `st_lt_memory.py`**: Updated `get_communication_history()` to iterate over list-valued message entries.
- **`test_llm_agent.py`**: Updated `test_apply_plan_adds_to_memory` assertion to match the new list-based action storage.
- **New**: `tests/test_memory/test_memory_staging.py` — 9 focused tests covering:
  - Single and multiple messages preserved in a list
  - Multiple actions preserved
  - Observations still overwrite (state-based)
  - Non-additive types still overwrite
  - Mixed types coexist correctly
  - `MemoryEntry.__str__` formats lists properly
  - `ShortTermMemory.get_communication_history` handles list messages

### Design Decision
Per the discussion in #76, communicative events (messages, actions) should be **additive** — an agent receiving five messages in one step should have all five recorded. Observations may remain state-based to prevent LLM context bloat.

The `ADDITIVE_EVENT_TYPES` set is a class attribute on `Memory`, making it easy for subclasses to extend with additional event types if needed.

### Test Plan
- [x] All 271 existing tests pass
- [x] 9 new tests for staging area behavior
- [x] Pre-commit hooks pass (ruff check, ruff format, codespell)

## Release Notes

* **New Features**
  * Messages and actions within a step now accumulate as lists, preserving all events instead of overwriting.
  * Added support for rendering lists in memory content with indexed enumeration.
  * New async API for memory updates.

* **Bug Fixes**
  * Improved communication history handling to properly display multiple messages per step.

PR #243, opened 2026-03-18
fix(memory): prevent add_to_memory() from overwriting events of the same type

### LINKED ISSUE #137: Memory staging area overwrites concurrent events of same type (leads to Data Loss)
**Describe the bug**

The base `Memory` class in mesa-llm uses a dictionary-based staging area `step_content` that only stores one entry per event type. When multiple events of the same type occur during a single simulation step (e.g., an agent receiving messages from multiple neighbors), subsequent calls to `add_to_memory` overwrite previous ones, leading to permanent data loss.

**Expected behavior**
In line with discussions at #76 

Communicative events (messages, actions, reasoning steps) should be additive. An agent receiving five messages in one step should have all five recorded.
Entries of `type=observations` may remain state-based (overwriting) to prevent LLM context bloat, but discrete events must be preserved.

**To Reproduce**
```
"""Sample test to show that step_content overwrites same-type entries"""
from mesa_llm.memory.st_memory import ShortTermMemory

agent = MagicMock()
agent.model.steps = 1

memory = ShortTermMemory(agent=agent, n=5)

# Simulate 3 agents sending messages to this agent in the same step
memory.add_to_memory("message", {"sender": "Agent_1", "msg": "Attack!"})
memory.add_to_memory("message", {"sender": "Agent_2", "msg": "Defend!"})
memory.add_to_memory("message", {"sender": "Agent_3", "msg": "Retreat!"})

print(f"Messages in step_content: {memory.step_content}")
# Output: {"message": {"sender": "Agent_3", "msg": "Retreat!"}}
# Agent_1 and Agent_2's messages are overwritten 

# Proof: only 1 message key exists
assert list(memory.step_content.keys()).count("message") == 1
print(
    f"\nOnly Agent_3's message survived. Agent_1 and Agent_2 lost.(length of step_content: {list(memory.step_content.keys()).count('message')})"
)
```
**Additional context**
The initial fix I had in mind was to transform the datatype of `step_content` to enable it to store additional lists for when multiple messages are sent during the same step.

But on further thought I think this would confuse the llm more and make it slower by supplying it a large amount of memory to parse through rather than providing it crisp info.

@sanika-n @wang-boyu @colinfrisch , What do you guys think about this?

### LINKED ISSUE #242: bug: add_to_memory() silently overwrites events of the same type within a single step
## Describe the bug
`add_to_memory()` in `mesa_llm/memory/memory.py` uses `self.step_content[type] = content` which overwrites any previous entry sharing the same type key. If an agent receives two messages or makes two tool calls in a single step, only the last event is retained and all prior ones are permanently lost.

## To Reproduce
Call `add_to_memory("message", {...})` twice in the same step and check `step_content`, only the second entry survives.
```python
memory.add_to_memory("message", {"content": "hello"})
memory.add_to_memory("message", {"content": "world"})
print(memory.step_content)  # {"message": {"content": "world"}} — WRONG, first message lost
```

## Expected behavior
All events of the same type within a step should be preserved.

## Root Cause
Dictionary key collision, same type key gets overwritten on each call.

## Fix
Change to `self.step_content.setdefault(type, []).append(content)`

I have a fix ready and will submit a PR immediately.

abikooo:
~~Fixes #242~~
Fixes #137

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -342,3 +342,12 @@ final_architecture.png
 GSoC_2025.md
 /architectures
 .instructions/
+
+# Diagnostic files
+output.txt
+pytest_failures.txt
+pytest_failures_all.txt
+show_stat.txt
+coverage.txt
+inbuilt_history.txt
+
--- a/mesa_llm/memory/episodic_memory.py
+++ b/mesa_llm/memory/episodic_memory.py
@@ -104,6 +104,7 @@ class EpisodicMemory(Memory):
         """
         Safely extracts importance score regardless of data structure.
         Handles:
+        - Nested in list: {"msg": [{"importance": 5}]}
         - Nested: {"msg": {"importance": 5}}
         - Flat:   {"importance": 5}
         """
@@ -112,7 +113,13 @@ class EpisodicMemory(Memory):
             return val if isinstance(val, (int, float)) else 1
 
         for value in entry.content.values():
-            if isinstance(value, dict) and "importance" in value:
+            if isinstance(value, list) and value:
+                # Handle unified structure where each type maps to a list
+                item = value[0]
+                if isinstance(item, dict) and "importance" in item:
+                    val = item["importance"]
+                    return val if isinstance(val, (int, float)) else 1
+            elif isinstance(value, dict) and "importance" in value:
                 val = value["importance"]
                 return val if isinstance(val, (int, float)) else 1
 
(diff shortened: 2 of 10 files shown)


souro26:
Please add the new tests in existing test files instead of making a new one

wang-boyu:
Thank you for working on this. Closing as superseded by #165, which has now been merged and closed #137.

PR #245, opened 2026-03-18
fix(llm_agent): prevent TypeError when agent.pos is None in _build_observation()

### LINKED ISSUE #244: bug: _build_observation() crashes with TypeError when agent.pos is None
## Describe the bug
In `mesa_llm/llm_agent.py`, the `_build_observation()` method calls `tuple(self.pos)` 
to query neighbors on the grid. If an agent has been instantiated but not yet placed 
on the grid, `self.pos` is `None`, causing the entire simulation to crash immediately.

## To Reproduce
```python
agent = LLMAgent(model=model, vision=1)
agent.pos = None  # agent not placed on grid yet
agent.generate_obs()  # TypeError: 'NoneType' object is not iterable
```

## Expected behavior
If an agent has no position yet, `_build_observation()` should safely return zero 
neighbors instead of crashing.

## Root Cause
Missing `None` guard before calling `tuple(self.pos)` in both the discrete grid 
and continuous space branches of `_build_observation()`.

## Fix
Add guard clauses:
```python
if self.pos is None:
    neighbors = []
```

I have a possible fix ready and will submit a PR immediately.

abikooo:
Fixes #244


wang-boyu:
Thanks for the fix. `main` is merged into this branch to solve conflicts. Merging now.

wang-boyu (approved):

PR #256, opened 2026-03-21
fix: add serialization-safe agent_id and to_dict() to MemoryEntry (#156)

### LINKED ISSUE #156: send_message/asend_message store Agent objects in memory content causing JSON serialization errors

aatmaj28:
Summary
Fixes #156 — MemoryEntry stores a direct LLMAgent object reference in its agent field, which causes TypeError: Object of type LLMAgent is not JSON serializable when memory entries are serialized.

Root Cause
The MemoryEntry dataclass (in mesa_llm/memory/memory.py) has agent: "LLMAgent" as a field. Every memory module (st_memory, lt_memory, st_lt_memory, episodic_memory) creates entries with agent=self.agent, embedding the full Agent object. Any downstream code that attempts json.dumps() on a MemoryEntry or its fields fails.

Changes
- Added agent_id: int | None = None field to MemoryEntry with __post_init__ that auto-populates from agent.unique_id
- Added to_dict() method that returns a JSON-serializable dictionary (content, step, agent_id)
- Fully backwards compatible: the existing agent field is unchanged, so all 9 call sites across 4 memory modules continue to work without modification

Testing
Existing tests unaffected — change is purely additive (new field with default + new method).

Context
I'm a GSoC 2026 candidate working on the Mesa-LLM stabilization project proposal. This is one of the critical issues I identified for Phase 1.

### DIFF
--- a/mesa_llm/memory/memory.py
+++ b/mesa_llm/memory/memory.py
@@ -15,11 +15,28 @@ if TYPE_CHECKING:
 class MemoryEntry:
     """
     A data structure that stores individual memory records with content, step number, and agent reference. Each entry includes `rich` formatting for display. Content is a nested dictionary of arbitrary depth containing the entry's information. Each entry is designed to hold all the information of a given step for an agent, but can also be used to store a single event if needed.
+
+    The agent field stores a weak reference to avoid circular references and
+    to keep entries JSON-friendly when the agent reference is not needed.
     """
 
     content: dict
     step: int | None
     agent: "LLMAgent"
+    agent_id: int | None = None
+
+    def __post_init__(self):
+        """Store the agent's unique_id for serialization-safe access."""
+        if self.agent is not None and self.agent_id is None:
+            self.agent_id = self.agent.unique_id
+
+    def to_dict(self) -> dict:
+        """Return a JSON-serializable representation of this memory entry."""
+        return {
+            "content": self.content,
+            "step": self.step,
+            "agent_id": self.agent_id,
+        }
 
     def __str__(self) -> str:
         """


IlamaranMagesh (commented):
Coul you please elaborate what's the issue here? I see the issue #156 is already closed and it has some fixes on JSON serialisation error.

wang-boyu:
Closing because #156 was already fixed, and this PR does not address that code path. Making `MemoryEntry` json-serializable could be a useful enhancement, but there is no current built-in logic that requires it. If that need comes up, we can revisit it in the future. Thanks for the effort though.

PR #258, opened 2026-03-22
improve RateLimitError message with actionable guidance

ospreyboi:
Problem:
When a user exceeds their free  API quota while running a mesa-llm simulation, this library raises a raw LiteLLM exception that contains almost 100-200 lines of JSON. This makes it very difficult for new users to understand what went wrong or what to do next. I ran into this issue myself while running the negotiation example with a free Gemini API key.
What I changed:
In module_llm.py, I wrapped the completion() call inside 'generate()' and the acompletion() call inside agenerate() with a try/except block. When a RateLimitError is caught, instead of letting the raw exception propagate, it re raises with a clean human-readable message that tells the user their quota is exceeded, suggests waiting a few minutes or switching to a different model, and provides a direct link to check their API quota.
Before this fix:
litellm.RateLimitError: geminiException - {"error": {"code": 429 ... 200 lines of JSON
After this fix:
RateLimitError: Rate limit exceeded for model 'gemini/gemini-2.0-flash'. 
You have exceeded your API quota. Please wait a few minutes and try again, 
or switch to a different model. 
To check your quota visit: https://ai.google.dev/gemini-api/docs/rate-limits
Note:
Adding a proper unit test for this fix requires addressing the testability of the retry decorator in module_llm.py , current tenacity retry setup causes tests to hang. I have raised this as a separate concern.

### DIFF
--- a/mesa_llm/module_llm.py
+++ b/mesa_llm/module_llm.py
@@ -104,6 +104,38 @@ class ModuleLLM:
 
         return messages
 
+    def _build_rate_limit_error(self, error: RateLimitError) -> RateLimitError:
+        provider = self.llm_model.split("/", 1)[0].lower()
+        docs_url = {
+            "anthropic": "https://platform.claude.com/docs/en/api/rate-limits",
+            "gemini": "https://ai.google.dev/gemini-api/docs/rate-limits",
+            "novita": "https://novita.ai/docs/guides/llm-rate-limits",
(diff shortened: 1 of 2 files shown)


IlamaranMagesh (commented):
Thank you but this catches error only in synchronous generation. There is an async method as well. Also, was it the same case for other errors?

wang-boyu (approved):

PR #262, opened 2026-03-24
fix: relative imports for examples in mesa-llm

EloZiko:
Updated the three examples (negotiation, epstein_civil_violence, and sugarscrap_g1mt) to use relative imports instead of absolute imports because of ModuleNotFoundError when running solara from the example directory as described in the README

Maybe it's better to change the README ? 

Tested locally by running solara run app.py from the example directory

### DIFF
--- a/examples/epstein_civil_violence/app.py
+++ b/examples/epstein_civil_violence/app.py
@@ -9,8 +9,8 @@ from mesa.visualization import (
     make_space_component,
 )
 
-from examples.epstein_civil_violence.agents import Citizen, CitizenState, Cop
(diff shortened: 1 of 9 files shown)


wang-boyu:
Thanks for reporting this. However, I can't reproduce this with the current codebase. Could you please share your exact steps to get the import errors? What I did was

1. create a fresh virtual environment and activate it
2. install mesa-llm: `pip install -e .`
3. go to an example, e.g., `cd examples/epstein_civil_violence`
4. run the model: `solara run app.py` and it runs without any issues

PR #277, opened 2026-03-30
Fix: coerce numeric arguments in ToolManager

BhoomiAgrawal12:
### Pre-PR Checklist
- [T] This PR is a bug fix, not a new feature or enhancement.

### Summary
Fixes a bug where `ToolManager` passed numeric arguments as strings, causing tools annotated with `int`/`float` to fail. This impacted tool execution for LLMs that return numbers as JSON strings (e.g., Ollama).

`ToolManager` promises automatic type coercion in its docstring, but `_process_tool_call()` only filtered args by name. When an LLM returned `"35.0"` (string) for a `float` parameter, the tool received a string and raised an error. Expected behavior is to coerce to numeric types before invocation.

### Bug / Issue
Fixes #276 

### Implementation
- Added coercion in `_process_tool_call()` for `float` and `int` parameters.
- Added two regression tests in `tests/test_tools/test_tool_manager.py`:
  - string → float (`"35.0"` -> `35.0`)
  - string → int (`"5"` -> `5`)
### Testing
- `pytest tests/test_tools/test_tool_manager.py -v`

### Additional Notes
- Coercion is intentionally limited to `int` and `float` to avoid unintended conversions.
- If coercion fails (invalid value), the original input is passed through so tools still surface meaningful errors.

### DIFF
--- a/mesa_llm/tools/inbuilt_tools.py
+++ b/mesa_llm/tools/inbuilt_tools.py
@@ -208,8 +208,14 @@ def speak_to(
         try:
             listener_agents_unique_ids = json.loads(listener_agents_unique_ids)
         except (json.JSONDecodeError, ValueError):
-            listener_agents_unique_ids = [int(x.strip()) for x in listener_agents_unique_ids.strip("[]").split(",") if x.strip()]
-    listener_agents_unique_ids = [int(uid) for uid in (listener_agents_unique_ids or [])]
+            listener_agents_unique_ids = [
+                int(x.strip())
+                for x in listener_agents_unique_ids.strip("[]").split(",")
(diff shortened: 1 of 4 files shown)


wang-boyu:
Thanks for reporting the issue and sending the fix. I made a small update to use `typing.get_type_hints()` when reading tool annotations, so string / deferred annotations like `"float"` resolve to the actual `float` type before coercion.

LGTM now. Merging. If this doesn't work for you, please feel free to re-open the issue.

wang-boyu (approved):

BhoomiAgrawal12:
Thanks for the addition and merge @wang-boyu !

PR #285, opened 2026-03-31
Fix: Correctly raised value error with descriptive error message in case of Invalid model instead of not a function calling model 

### LINKED ISSUE #284: [Bug] Incorrect Error Raised When Model Name is Invalid
**Describe the bug**

When Given a invalid model name it is show that litellmError not a function calling model
It will confuse thinking their model is correct but it's not function calling

**Expected behavior**

When Invalid model name is give we should raise Value Error along with model name and api_base for better debugging on user side along with litellm actual error message 

**To Reproduce**

llm = ModuleLLM(model="openai/get-4o")

Initialize llm with this model or any invalid you get an not a function calling model

DineshThumma9:
> [!NOTE]

### Pre-PR Checklist
- [x] This PR is a bug fix, not a new feature or enhancement.

### Summary

 When given an invalid model name it throws Error with message not a function calling model it might trick user in thinking their model is valid but not function calling and does'nt address actual model invalid issue 

### Bug / Issue

Fix : #284 

### Implementation

I added an model validity check right after api key check i used 

```
litellm.get_model_info() 
```

wrapped in try and except block and raise ValueError when caught an error and show a descriptive message

### DIFF
--- a/mesa_llm/module_llm.py
+++ b/mesa_llm/module_llm.py
@@ -5,6 +5,7 @@ from dotenv import load_dotenv
 from litellm import acompletion, completion, litellm
 from litellm.exceptions import (
     APIConnectionError,
+    NotFoundError,
     RateLimitError,
     Timeout,
 )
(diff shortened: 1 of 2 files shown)


wang-boyu:
Thanks for the fix @DineshThumma9 . I think `litellm.get_model_info()` is too strict to use in `__init__`, because it also raises for unmapped but valid models such as custom deployments.

I updated the logic so that initialization no longer mislabels invalid models as lacking function-calling support, and invalid-model errors are raised later at `completion()` time when LiteLLM actually fails to call the model.

Before I merge it, could you please verify that this works for you as expected? Thanks.

DineshThumma9:
Thanks @wang-boyu! I ran it locally too and it works as expected. No problems on my end. (Also, apologies—I closed this by mistake a moment ago!)

wang-boyu:
No worries at all, and thanks for the quick confirmation. Merging this now.

wang-boyu (approved):

PR #286, opened 2026-04-01
fix: change executor tool_choice from required to auto so reasoning respects no-action decisions

saakshisinghal14:
Fixes #253

## Summary

When CoT reasoning concludes that no action should be taken (e.g., "I choose NOT to call adopt_opinion"), the executor phase still forces a tool call because `tool_choice="required"`. This causes the LLM to pick an arbitrary tool — often an unrelated one like `move_one_step` — contradicting its own reasoning.

The same issue exists in the base `Reasoning.execute_tool_call()` and `aexecute_tool_call()` methods used by ReAct and ReWOO.

## Changes

- Changed `tool_choice="required"` to `tool_choice="auto"` in all four executor call sites (CoT sync/async, base Reasoning sync/async)
- Updated the executor system prompt to explicitly instruct the LLM that it may skip tool calls when the plan concludes no action is needed
- Updated existing test assertion to match the new `tool_choice` value

## Files changed

- `mesa_llm/reasoning/cot.py` — `plan()` and `aplan()` executor phases
- `mesa_llm/reasoning/reasoning.py` — `execute_tool_call()` and `aexecute_tool_call()`
- `tests/test_reasoning/test_reasoning.py` — updated assertion for `tool_choice`

## Tests

All 21 reasoning tests pass:

### DIFF
--- a/mesa_llm/reasoning/cot.py
+++ b/mesa_llm/reasoning/cot.py
@@ -130,12 +130,15 @@ class CoTReasoning(Reasoning):
         # Pass plan content to agent for display
         if hasattr(self.agent, "_step_display_data"):
             self.agent._step_display_data["plan_content"] = chaining_message
-        system_prompt = "You are an executor that executes the plan given to you in the prompt through tool calls."
+        system_prompt = (
+            "You are an executor that executes the plan given to you in the prompt through tool calls. "
+            "If the plan concludes that no action should be taken, do not call any tool."
+        )
         llm.system_prompt = system_prompt
         rsp = llm.generate(
             prompt=chaining_message,
             tool_schema=self.agent.tool_manager.get_all_tools_schema(selected_tools),
-            tool_choice="required",
+            tool_choice="auto",
         )
         response_message = rsp.choices[0].message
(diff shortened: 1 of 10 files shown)


wang-boyu:
Thanks for the contribution and the discussion here. I pushed some follow-up commits / fixes:

- fixed the ReWOO no-op case when `llm_plan.tool_calls` is `None`
- added a `tool_calls` parameter so execution-side `tool_choice` is configurable (default to `"auto"`) while planning still uses `tool_choice="none"`
- refactored CoT to use the shared `execute_tool_call()` path to align with ReAct and ReWOO
- aligned memory handling across CoT, ReAct, and ReWOO to have consistent `plan_execution` logging
- updated some docstrings and test cases
- merged `main` branch to resolve conflicts

Thanks again for the contribution. I'm merging this now.

wang-boyu (approved):

saakshisinghal14:
Thanks @wang-boyu  for the detailed follow-up changes and improvements!

I really like the alignment across CoT/ReAct/ReWOO and the configurable tool_calls design — makes the execution flow much cleaner and more consistent.

Glad I could contribute to this fix. Looking forward to building on this further!

PR #289, opened 2026-04-05
fix: add include_default_tools parameter to LLMAgent to prevent unwanted inbuilt tool calls

saakshisinghal14:
Fixes #254

## Summary

When an agent registers only domain-specific tools (e.g. `adopt_opinion`), the LLM executor can still call inbuilt tools like `move_one_step` — even when the simulation has nothing to do with movement.

This happens because `ToolManager` auto-registers inbuilt tools globally, with no way to opt out. When the executor must choose a tool, it may select these unintended defaults.

## Changes

- Added `INBUILT_TOOL_NAMES` (frozenset) in `inbuilt_tools.py` to centrally define default tool names
- Added `include_default_tools: bool = True` parameter to `LLMAgent.__init__()`
  - `True` (default): preserves existing behavior (fully backward compatible)
  - `False`: filters out inbuilt tools (`move_one_step`, `teleport_to_location`, `speak_to`) from the agent’s `ToolManager`, leaving only user-registered tools
- Added tests covering:
  - default behavior (inbuilt tools present)
  - opt-out behavior (inbuilt tools removed)
  - preservation of user-defined tools

## Implementation Note

Filtering is applied after tool registration, ensuring compatibility with the existing tool registration flow while allowing selective removal of default tools per agent.

## Usage

```python
agent = LLMAgent(
    model=model,
    reasoning=CoTReasoning,
    include_default_tools=False,  # only domain tools visible to LLM
)

### DIFF
--- a/mesa_llm/llm_agent.py
+++ b/mesa_llm/llm_agent.py
@@ -17,6 +17,7 @@ from mesa_llm.reasoning.reasoning import (
     Observation,
     Reasoning,
 )
+from mesa_llm.tools.inbuilt_tools import INBUILT_TOOL_NAMES
 from mesa_llm.tools.tool_manager import ToolManager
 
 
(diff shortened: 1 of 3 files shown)


saakshisinghal14:
Thanks for pointing this out @BhoomiAgrawal12 . I wasn't aware of PR #167 when I opened this — it predates my involvement with the repo.

Looking at it, #167 takes a more thorough approach by adding the flag to ToolManager itself and including a remove_tool() method, which is cleaner than my post-creation filtering. I'm happy to defer to #167 if the maintainers prefer that approach.

PR #290, opened 2026-04-06
fix: replace deprecated seed= with rng= in all example models

ospreyboi:
Fixes part of #273
All three example models (negotiation, sugarscrap_g1mt, epstein_civil_violence) were using the deprecated seed= parameter in both model.py and app.py files. This causes it to create a FutureWarning in Mesa 3.5 and will break in Mesa 4.0.
This  replaces all occurrences of seed=None with rng=None in the function signatures and super().__init__(seed=seed) with super().__init__(rng=rng) in all three examples.
All 285 tests pass after the changes made.

### DIFF
--- a/examples/epstein_civil_violence/app.py
+++ b/examples/epstein_civil_violence/app.py
@@ -38,7 +38,7 @@ agent_colors = {
 }
 
 model_params = {
-    "seed": {
+    "rng": {
         "type": "InputText",
         "value": 42,
         "label": "Random Seed",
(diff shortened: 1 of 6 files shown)


wang-boyu:
Thanks for sending the PR. I'd prefer to have it as part of #273 when we migrate to Mesa 4. Currently mesa-llm has mesa pinned <4 (#272).

Another reason is that switching examples to `rng=` causes a Solara seed-reset issue on Mesa 3.x. It is fixed in Mesa 4 (see https://github.com/mesa/mesa/pull/3518) but has not been backported to any 3.x release yet. If a future Mesa 3.x release backports this fix, we can revisit merging this earlier before #273, since it's also compatible with Mesa 3.

ospreyboi:
@wang-boyu 
Thank you for the explanation. I understand and I'll keep this PR open and link it to #273 for when the Mesa 4 migration is completed would be happy to update it then if needed.

PR #298, opened 2026-04-17
fix(llm_agent): stop redundant system_prompt in observations

### LINKED ISSUE #297: bug: redundant system prompt in observation context
### **Describe the bug** 
In `mesa_llm/llm_agent.py`, the `_build_observation(`) method includes the entire `self.system_prompt` in the `self_state` dictionary. Since observations are committed to memory every step, this causes the static system prompt to be duplicated in short-term memory repeatedly, causing exponential token waste.

### **To Reproduce** 
Run a basic simulation for 5 steps with a long system prompt. Inspecting `agent.memory.step_content` will show the prompt string present in every observation entry.

### **Expected behavior** 
Observation memory should strictly track dynamic state changes. Static configuration data like the system prompt should not be persisted redundantly in agent memory

### **Root Cause** 
`self.system_prompt` is explicitly included in the `self_state` dictionary within `_build_observation()`.Fix Remove `system_prompt `from the observation dictionary.

**I have a fix ready and will submit a PR immediately.**

abikooo:
Fixes #297

### DIFF
--- a/tests/test_llm_agent.py
+++ b/tests/test_llm_agent.py
@@ -1159,3 +1159,30 @@ def test_system_prompt_proxies_llm_prompt(basic_agent):
     basic_agent.llm.system_prompt = "LLM prompt"
 
     assert basic_agent.system_prompt == "LLM prompt"
+
+
+def test_observation_excludes_system_prompt(monkeypatch):
+    """ensure system_prompt is NOT included in the observation state to save tokens."""
+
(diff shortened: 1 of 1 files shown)


wang-boyu:
I don't see a code change here beyond an added test. On current `main`, the issue seems to be already fixed.

Since this PR doesn't change any implementation, I’m closing it as redundant.

PR #305, opened 2026-04-23
fix: raise TypeError for invalid prompt types in _build_messages()

### LINKED ISSUE #304: bug problem : _build_messages() which silently ignores invalid prompt types instead of raising TypeError
When someone puts an invalid type like an integer is passed to _build_messages(), 
the function is silently ignoring it and returns only the system message 
with no error or warning.

Steps to reproduce:
import os
os.environ['OPENAI_API_KEY'] = 'test_key'
from mesa_llm.module_llm import ModuleLLM
m = ModuleLLM('openai/gpt-4o')
print(m._build_messages(123))

Output: [{'role': 'system', 'content': ''}]
Expected: TypeError with a clear message about invalid prompt type

The type hint says str  list[str] None but there is no validation.
Invalid types pass through silently which can cause confusing LLM behavior.

Proposed fix: add an else clause that raises TypeError

ospreyboi:
Fixes #304

When an invalid type like an integer is put inside  to _build_messages(), 
the function was tending to  silently ignoring the issue  and returning only the system 
message with no error or warning. This could cause confusing LLM 
behavior which would have been hard to debug.

This PR adds validation ,if the prompt type is not str, list[str], 
or None, a TypeError is raised with a clear message telling the user 
what type was passed and what types are expected.

Also changed `if prompt:` to `if prompt is not None:` so that empty 
strings are handled properly.

All 381 tests were passing.

### DIFF
--- a/mesa_llm/module_llm.py
+++ b/mesa_llm/module_llm.py
@@ -126,6 +126,20 @@ class ModuleLLM:
             if isinstance(prompt, str):
                 messages.append({"role": "user", "content": prompt})
             elif isinstance(prompt, list):
+                invalid_prompt = next(
+                    (
+                        (index, value)
+                        for index, value in enumerate(prompt)
+                        if not isinstance(value, str)
+                    ),
+                    None,
+                )
+                if invalid_prompt is not None:
+                    index, value = invalid_prompt
+                    raise TypeError(
+                        f"Invalid prompt list element at index {index}: "
+                        f"type '{type(value).__name__}'. Expected str."
+                    )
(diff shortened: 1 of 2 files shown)


wang-boyu:
Thanks for the fix. I added an additional check for list elements so that `prompt=[123]` also raises a `TypeError`.

Merging now.

wang-boyu (approved):

ospreyboi:
@wang-boyu thanks for the review and the checks 

Appreciate the merge!

